PR #73, opened 2022-07-22
create a tentative protocol to switch desktop modes

Aki-7:
## [293d92b](https://github.com/zigen-project/zen/pull/73/commits/293d92b96c3451c24ada5fb86a7db6bd65509306)

- Create a tentative protocol to switch desktop modes
- Add test code to validate protocol xml

### DIFF
--- a/tests/test-runner.c
+++ b/tests/test-runner.c
@@ -1,0 +1,102 @@
+#include "test-runner.h"
+
+#include <assert.h>
+#include <errno.h>
+#include <signal.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+extern const struct test __start_test_section, __stop_test_section;
+
+static int is_atty = 0;
+
+#define RED "\033[31m"
+#define GREEN "\033[32m"
+
+static void
+stderr_set_color(const char *color)
+{
+  if (is_atty) fprintf(stderr, "%s", color);
+}
+
+static void
+stderr_reset_color(void)
+{
+  if (is_atty) fprintf(stderr, "\033[0m");
+}
+
+static void
+run_test(const struct test *t)
+{
+  t->run();
+
+  exit(EXIT_SUCCESS);
+}
+
+int
+main()
+{
+  const struct test *t;
+  pid_t pid;
+  int total, pass;
+  siginfo_t info;
+
+  if (isatty(fileno(stderr))) is_atty = 1;
+
+  pass = 0;
+  for (t = &__start_test_section; t < &__stop_test_section; t++) {
+    int success = 0;
+
+    pid = fork();
+    assert(pid >= 0);
+
+    if (pid == 0) run_test(t); /* never returns */
+
+    if (waitid(P_PID, pid, &info, WEXITED)) {
+      stderr_set_color(RED);
+      fprintf(stderr, "waitid failed: %s\n", strerror(errno));
+      stderr_reset_color();
+
+      abort();
+    }
+
+    switch (info.si_code) {
+      case CLD_EXITED:
+        if (info.si_status == EXIT_SUCCESS)
+          success = !t->must_fail;
+        else
+          success = t->must_fail;
+
+        stderr_set_color(success ? GREEN : RED);
+        fprintf(
+            stderr, "test \"%s\": \texit status %d", t->name, info.si_status);
+        break;
+
+      case CLD_KILLED:
+      case CLD_DUMPED:
+        if (t->must_fail) success = 1;
+
+        stderr_set_color(success ? GREEN : RED);
+        fprintf(stderr, "test \"%s\":\tsignal %d", t->name, info.si_status);
+        break;
+    }
+
+    if (success) {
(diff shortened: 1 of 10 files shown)


tests/validate-xml-test.c
```diff
@@ -0,0 +1,43 @@
+#include <assert.h>
+#include <libxml/parser.h>
+
+#include "test-runner.h"
+
```
liwii, line 5: I think it would be better to have the description of the test (as a comment).

tests/test-runner.c
```diff
@@ -0,0 +1,102 @@
+#include "test-runner.h"
```
liwii, line 1: Similar to the comment to tests/validate-xml-test.c: it would be helpful for future contributors to leave a comment which describes what the file is for.

tests/validate-xml-test.c
```diff
@@ -0,0 +1,43 @@
+#include <assert.h>
```
liwii, line 1: I would create another directory `tests/xml` and put this file into it (at least I think individual test cases should be put in different directories from that of the test runner.)

liwii (changes requested):

tests/validate-xml-test.c
Aki-7: Thanks! That's a good idea!

tests/validate-xml-test.c
Aki-7: I know there are different opinions on how much comment to put in the code. This is not to say that I disagree with you, but as an OSS project we just want to align our perceptions with those of the developers. And I think it is best if the behavior can be understood by function names, file names, and simple code, with as few comments as possible.

Also, writing comments takes a certain amount of time and effort, and I want the contributor to focus on writing code that is as simple and easy to understand as possible, and I do not want to force the contributor to write comments.

If the behavior of this file or the function is not clear to you, I would be very happy if you could point it out to me, but I would appreciate it if you could tell me which part of the file is difficult to understand. I will then rewrite the code to be more understandable or even write a comment if necessary.

(-> see the next comment)

tests/test-runner.c
Aki-7: For example, about this `TEST(validate_zigen_xml)`, before that, it validates zen-desktop.xml, not zigen.xml, so the name `TEST(validate_zigen_xml)` is correct, which I will fix.

So, if I add a comment to `TEST(validate_zen_desktop_xml)`, it would be something like "check if `zen-desktop.xml` is written in the correct format", which I think you can easily see from the code. I think. Or if you say that the function `validate_xml` should be named `validate_xml_with_wayland_dtd` because it validates xml with wayland.dtd, I would agree.

Again, this is just to align perceptions and I would like to hear your opinions.
@zigen-project/zigen-developers

tests/validate-xml-test.c
liwii: Thanks for the reply. I totally agree that there are a lot of advantages to leaving as few comments as possible. \

Let me clarify what I want you to add to this file. To me,  what "validating XML" means seems very ambiguous. We can tell that by reading the whole code (it is testing whether the libxml functions can parse the zigen xmls, right?), I think it would be better to be able to get this kind of high-level picture more easily. Especially, when writing a test, this kind of description should be in the developer's mind in advance. For this purpose, writing a one-line comment should be easier than having a longer function name (`validate_zigen_xml_can_be_parsed_with_...`)

tests/test-runner.c
liwii: For this file, I requested comments from a different perspective: this will be the common infrastructure for all the tests in this repository, so I think at least writing a how-to-use comment will be helpful for future developers. They do not need to understand what the test harness is actually doing. A simple comment (or some document or help) will save a lot of time.

tests/validate-xml-test.c
Aki-7: > To me, what "validating XML" means seems very ambiguous.
You are absolutely right.

In this function, I validate the XML using a DTD file provided by wayland.
dtd: https://wikipedia.org/wiki/Document_Type_Definition
wayland.dtd: https://gitlab.freedesktop.org/wayland/wayland/-/blob/main/protocol/wayland.dtd

I fixed the function name in [3a913cf](https://github.com/zigen-project/zen/pull/73/commits/3a913cf91389cf89c8d942a9c8e2effcfa521632).
What do you think?

tests/test-runner.c
Aki-7: I was very convinced. Thanks!

tests/validate-xml-test.c
liwii: LGTM

tests/test-runner.c
Aki-7: done!

Aki-7:
@rtr-dnd 
I would like you to look over the protocol once if possible.

rtr-dnd:
@Aki-7 Thanks for suggestion. I’ve read the xml and it was very easy to understand. Perhaps I should read other protocol xmls too.

rtr-dnd (approved):

liwii (approved):

Issue #74, opened 2022-07-23
File names are output as relative paths in log.


Aki-7:
Currently, this is what the zen's log looks like,

> 00:00:09.644 [../zen/main.c:47] Caught signal 2

But it should be like,

> 00:00:09.644 [zen/main.c:47] Caught signal 2

The sway logger would be very helpful in implementing this.

Aki-7:
closed by #75.
Thanks @liwii!!!

PR #75, opened 2022-07-24
Remove relative path prefix from __FILE__

liwii:
# Issue
#74 

# Summary
Remove prefix from `__FILE__` macro with a compiler option.
 See `-fmacro-prefix-map` in the [gcc doumentation](https://gcc.gnu.org/onlinedocs/gcc/Preprocessor-Options.html). In [the sway change I referred to](https://github.com/swaywm/sway/commit/2dc4978d8af326c310057ca8fd22a4c7f5d09335), it also handles the cases when the option is not available. However, the change adds another macro and complicates the program, and the option has been available for a long time (from gcc8 and clang10), so I decided to just ignore such a case.

# Test
I bulit `zen-desktop`, ran it, and saw the following output as the log.
```
00:00:00.026 [zen/server.c:255] WAYLAND_DISPLAY=wayland-1
00:00:00.026 [zen/server.c:256] XDG_RUNTIME_DIR=/run/user/1000
00:00:00.026 [zen/server.c:153] DISPLAY=:2
```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -30,7 +30,7 @@ build_root = meson.global_build_root().split('/')
 slen = source_root.length()
 blen = build_root.length()
 
-shorter_path_len = source_root.length() >= build_root.length() ? build_root.length() : source_root.length()
+shorter_path_len = slen >= blen ? blen : slen
 
 relative_source_parts = []
 common_prefix_len = 0
--- a/meson.build
+++ b/meson.build
@@ -35,26 +35,22 @@ shorter_path_len = slen >= blen ? blen : slen
 relative_source_parts = []
 common_prefix_len = 0
 
-foreach _ : build_root
-  if shorter_path_len > common_prefix_len and source_root[common_prefix_len] == build_root[common_prefix_len]
-    common_prefix_len += 1
-  else
-    break
-  endif
-endforeach
-
 i = 0
-foreach _ : build_root
-  if i >= common_prefix_len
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -22,6 +22,54 @@ foreach arg : global_args_maybe
...
+  i += 1
+endforeach
+
+i = 0
+foreach dir : source_root
+  if i >= common_prefix_len
+    relative_source_parts += dir
+  endif
+  i += 1
+endforeach
```
Aki-7, line 60: I think sway-like implementation is simple, but what do you think?
```suggestion
i = 0
in_prefix = true
foreach p : build_root
  if not in_prefix or p != source_root[i]
    in_prefix = false
    relative_dir_parts += '..'
  endif
  i += 1
endforeach

i = 0
in_prefix = true
foreach p : source_root
  if not in_prefix or build_root[i] != p
    in_prefix = false
    relative_dir_parts += p
  endif
  i += 1
endforeach
```

Aki-7 (commented):
Implementation is not the sway's way you are referring to. Is this assuming that the build directory is not in the source directory?

rtr-dnd (approved):

meson.build
liwii: Yup I agree, let me take it in... Thanks!!

Aki-7 (approved):
LGTM!

PR #76, opened 2022-07-24
Handle display system switch protocol

Aki-7:
### server
- handle zen_display_system switch request, and just return zen_display_system_applied event.
- refactor build system regarding protocols
  - I created a client, and the current build system does not work well with building clients. 

### client
- create a CUI client that send requests and receive events regarding zen_display_system.
  - This client application are to be remove when corresponding GUI application is developed

## How to check behavior

1. start `zen-desktop`

2. start a CUI client

```
$ ./build/cui-client/zen-desktop-cui-client
```

3. start another client in another shell

```
$ ./build/cui-client/zen-desktop-cui-client
```

4. enter a command, "screen" or "immersive" in a client

The client send zen_display_system::switch_type request with type argument, then both clients will receive zen_display_system::applied event.

### DIFF
--- a/include/zen/display-system.h
+++ b/include/zen/display-system.h
@@ -1,0 +1,26 @@
+#ifndef ZEN_DISPLAY_SYSTEM_H
+#define ZEN_DISPLAY_SYSTEM_H
+
+#include <wayland-server-core.h>
+#include <zen-desktop-protocol.h>
+
+struct zn_display_system_switch_event {
+  enum zen_display_system_type type;
+};
+
+struct zn_display_system {
+  struct wl_global* global;
+
+  struct wl_list resources;
+
+  struct wl_signal switch_signal;  // (struct zn_display_system_switch_event *)
+};
+
+void zn_display_system_send_applied(
+    struct zn_display_system* self, enum zen_display_system_type type);
+
+struct zn_display_system* zn_display_system_create(struct wl_display* display);
+
+void zn_display_system_destroy(struct zn_display_system* self);
+
+#endif  //  ZEN_DISPLAY_SYSTEM_H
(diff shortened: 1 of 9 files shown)


zen/server.c
```diff
@@ -24,13 +26,13 @@ struct zn_server {
...
   struct wlr_allocator *allocator;
+  struct wlr_xwayland *xwayland;
 
   // these objects will be automatically destroyed when wl_display is destroyed
   struct wlr_compositor *w_compositor;
   struct wlr_xdg_shell *xdg_shell;
 
-  struct wlr_xwayland *xwayland;
-
+  struct zn_display_system *display_system;
```
shierote, line 35: For future reference, could you tell me how the order of the struct members here is determined?
I am guessing wlr objects in front, zn objects behind, etc.

include/zen/display-system.h
```diff
@@ -0,0 +1,26 @@
...
+#include <zen-desktop-protocol.h>
+
+struct zn_display_system_switch_event {
+  enum zen_display_system_type type;
+};
+
+struct zn_display_system {
+  struct wl_global* global;
+
+  struct wl_list resources;
```
shierote, line 14:
```suggestion
  struct wl_list resources; // wl_resource::link
```
(It's almost obvious.)

shierote (commented):
For compositor codes, LGTM.
If I have the time and energy, I will look at the client-side code after tomorrow.

zen/server.c
Aki-7: That's what I mean, but not strict rule, I believe.

include/zen/display-system.h
Aki-7: I don't think there is any confusion without comment.

Aki-7:
Thanks! Please take care of your self.

rtr-dnd (approved):
Confirmed that the new codes are working as expected in my env

Aki-7:
Thanks for your review!

PR #77, opened 2022-07-24
Add template for new feature pull request

Aki-7:
Little by little, each developer started doing different tasks. I know you are busy with your own development, but I think it is important to catch up with what other developers are developing. So I'm going to set up a section in the pull request so that other people can easily check the new behavior you made. What do you think?

Inspired by #75. 
#76 is an example.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/new_feature.md
+++ b/.github/PULL_REQUEST_TEMPLATE/new_feature.md
@@ -1,0 +1,15 @@
+## Summary
+
+<!--
+Please write what you did.
+If you have related issues, pull requests, etc., please post those as well.
+If you have made changes to the appearance, it would be helpful to include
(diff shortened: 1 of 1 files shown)


shierote (approved):
it looks simple and nice!

We may add the following items.
- Is there a related issue?
- What has not been done yet and needs to be done in the future.
- issue template (maybe as another pr)

liwii:
Having "Related issue" will help the reviewers understand the problem, so sounds good. Maybe the title should be "Related issue (if exists)"?

Aki-7:
Thanks! I agree with adding a few more items!

However, each of these may be open to further discussion, so I'll leave them out of the scope of this pull request.

I will file a new pull request for related issues section right after this, but I also welcome PRs from you to add other items.

Issue #78, opened 2022-07-24
Switching display system


Aki-7:
Enable switching between "immersive display system" used when working in XR and "screen display system" used when working on a physical monitor.
The details of what is rendered on each display system are not the scope of this issue, but running the rendering and event loop on the immersive display system is within the scope of this issue.

# Milestone
- [x] write a protocol to switch display systems. #73 
- [x] create a client which implement the protocol. #76 
- [x] create `openvr` backend as the first option of "immersive backend", and check if immersive display system is now available or not when a client request server to switch display system to "immersive". #84 
- [ ] run openvr rendering and event loop together with wayland event loop.
  - [ ] without any actual rendering #92
  - [ ] render some object, e.g. floor

PR #84, opened 2022-07-26
Immersive backend

Aki-7:
## Context

See #78

> create openvr backend as the first option of "immersive backend", and check if immersive display system is now available or not when a client request server to switch display system to "immersive".

## Summary

- add concept of immersive backend, which handles immersive devices like HMD.
- create openvr immersive backend as the first option of "immersive backend", and check if immersive display system is now available or not when a client request server to switch display system to "immersive".

## How to check behavior

1. build
2. start zen-desktop without startup command. In its log, DISPLAY environment variable to use is printed.
3. start alvr
```
$ DISPLAY=:0 alvr_launcher
```
4. start cui-client (see #76 for detail)
5. enter "immersive" command in the cui client.
6. you will soon receive applied event if steamvr is ready.
7. if you enter "immerisive" command without starting alvr, you will soon get warning which says no immersive device found.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -18,7 +18,8 @@ jobs:
         run: |
           sudo apt-get update &&
           sudo apt-get install \
-            libwayland-dev libwlroots-dev libpixman-1-dev libxml2-dev weston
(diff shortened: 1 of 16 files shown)


Aki-7:
no reaction :_(

Issue #90, opened 2022-07-29
basic keyboard input


watasuke102:
exit immediately when user press any key

watasuke102:
close by #91

PR #91, opened 2022-07-29
add zn_keyboard

watasuke102:
## Context

See: #90 

## Summary

When exiting zen, we have to run `pkill zen-desktop` until now.

By this PR, we can exit zen with pressing any key.

## How to check behavior

1. launch zen
2. press any key on the keyboard
3. terminate - with `exit(0)`


Aki-7:
As I read your code, there is only one zn_keyboard even if there are more than two physical keyboard connected, which means zn_keyboard is very different from sway_keyboard of sway compositor. A sway_keyboard is equivalent to one physical keyboard.
Is this as you intended?

zen/keyboard.c
```diff
@@ -0,0 +1,44 @@
...
+
+err:
+  return NULL;
+}
+
+void
+zn_keyboard_destroy(struct zn_keyboard* self)
+{
+  free(self);
+}
```
Aki-7, line 44: A new line is required.
Most editors have the feature to automatically insert a line break at the end of the file.

zen/keyboard.c
```diff
@@ -0,0 +1,44 @@
...
+
+void
+zn_keyboard_configure(
+    struct zn_keyboard* self, struct zn_input_device* input_device)
+{
+  self->key_listener.notify = handle_key;
+  wl_signal_add(
+      &input_device->wlr_input->keyboard->events.key, &self->key_listener);
+}
```
Aki-7, line 22: Consider the case where multiple keyboards are connected.

There are at least two problem here, 
1. wl_signal_add adds a listener to the signal's list. self->key_listener can be already inserted to another signal's list, and inserting already inserted element to a list is not safe in wayland's implementation. 
2. Only the last keyboard connected will trigger handle_key event.

Aki-7:
It may need to be changed from the design. Then I will review it again.

watasuke102:
I forgot to support multi device, and have implemented now.

include/zen/seat.h
```diff
@@ -3,11 +3,15 @@
...
-#include "zen/input-device.h"
+#include "zen/keyboard.h"
 
 #define ZEN_DEFAULT_SEAT "seat0"
 
-struct zn_seat;
+struct zn_seat {
+  struct wlr_seat* wlr_seat;
+  struct wl_list devices;    // zn_input_device::link
+  struct wl_list keyboards;  // zn_keyboard::link
```
Aki-7, line 13: `keyboards` is not actually used

zen/keyboard.c
```diff
@@ -0,0 +1,48 @@
...
+#include <wlr/types/wlr_seat.h>
+
+#include "zen-common.h"
+
+static void
+handle_key(struct wl_listener* listener, void* data)
+{
+  UNUSED(listener);
+  UNUSED(data);
+  exit(0);
```
Aki-7, line 12:
```suggestion

  // Terminate the program with a keyboard event for development convenience.
  exit(0);
```

Aki-7 (approved):
Good work!

Issue #83, opened 2022-07-26
Segmentation fault when stopping compositor


Aki-7:
## Problem

Segmentation fault occurs when stopping compositor.

## Cause

Deleting order of seat and input device seems not good.

PR #67, opened 2022-07-18
add input manager

watasuke102:
add input manager

PR #93, opened 2022-08-01
Exit normally

Aki-7:
## Context

See #83 

### 6e4d470d0f712e7067bda12c27818efe077662e5

Destroy zn_input_device before zn_seat

### 37a75d26dce176bb68753e96dee5295540f107c2

Exit more gracefully when key input is detected. If just  terminate with `exit(0)`, developers cannot know the termination process works well or not.

### 882d26b5563c1a91a989d289f5868ab11dd51f79

Exit with only Q key pressed. I want to use some key to switch keyboard between multiple computers.

## How to check behavior

1. build
2. start `zen-desktop`
3. press 'q'

You'll see `[zen/main.c:195] Exited gracefully` at the end of the log.

### DIFF
--- a/zen/server.c
+++ b/zen/server.c
@@ -159,8 +159,10 @@ zn_server_create(struct wl_display *display)
   char socket_name_candidate[16];
   char *xdg;
 
-  zn_assert(
-      server_singleton == NULL, "Tried to create zn_server multiple times");
+  if (!zn_assert(server_singleton == NULL,
+          "Tried to create zn_server multiple times")) {
+    goto err;
+  }
 
   self = zalloc(sizeof *self);
   if (self == NULL) {
--- a/zen/server.c
+++ b/zen/server.c
@@ -170,6 +170,7 @@ zn_server_create(struct wl_display *display)
     goto err;
   }
 
+  server_singleton = self;
   self->display = display;
   self->exit_code = EXIT_FAILURE;
   self->loop = wl_display_get_event_loop(display);
@@ -267,7 +268,6 @@ zn_server_create(struct wl_display *display)
   wl_signal_add(&self->display_system->switch_signal,
       &self->display_system_switch_listener);
 
-  server_singleton = self;
   return self;
 
 err_socket:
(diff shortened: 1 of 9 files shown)


Aki-7:
@watasuke102
Your PR #67 is so relevant to this. I want you to review this.

shierote (commented):
I think you need to rewrite `README.adoc` because in this doc there is instruction for stopping zen!

zen/server.c
```diff
@@ -149,6 +159,9 @@ zn_server_create(struct wl_display *display)
   char socket_name_candidate[16];
   char *xdg;
 
+  zn_assert(
+      server_singleton == NULL, "Tried to create zn_server multiple times");
```
shierote, line 163:
```suggestion
      server_singleton != NULL, "Tried to create zn_server multiple times");
```
I could be wrong, but isn't it like this?

watasuke102: `zn_assert` is check first argument, and if it is **false**, show message and return false.
In this case, it expects this is the first time that creating zn_server, so `server_singleton` must be NULL. So, it looks correct.

shierote: that's right...
I misunderstood. Thx!

watasuke102, line 163: if `server_singleton` is non-NULL, `zn_assert` returns false, but creation of zn_server will not stop.
This case, I think it should return there.

zen/server.c
```diff
@@ -285,6 +299,7 @@ zn_server_create(struct wl_display *display)
 void
 zn_server_destroy(struct zn_server *self)
 {
+  server_singleton = NULL;
```
watasuke102, line 302: Maybe I'm overthinking about this, but is there no prospect to be required `server_singleton` by destroy process in the future? In that case, it must fail despite zn_server is still aliving.
Also, this is my own preference, `free()` and assigning NULL should be placed near by.
Therefore, I think moving assignment of NULL to `server_singleton` after `free(self)` is works well.

watasuke102 (changes requested):

zen/server.c
Aki-7: You said it. Nice catch!

watasuke102 (approved):

Issue #85, opened 2022-07-28
func() or func(void)


Aki-7:
Now we have a two types of no-arg function definition, func() and func(void).

Whichever we chose, we should use only one of them.

PR #94, opened 2022-08-01
use func(void)

Aki-7:
## Context

#85 

## Summary

For definitions and declarations of functions without argument in C, I proposed to use
```c
void func(void);
```
instead of
```c
void func();
```

This should be applied to C file and header files which can be included by C files.
Not applied to C++ code and header files only for C++

## Reason

Each types of declaration/definition means different things.
Proposed one is more restriced.

This might be useful to understand the difference.
https://softwareengineering.stackexchange.com/questions/286490/what-is-the-difference-between-function-and-functionvoid

## How to check behavior

None.

### DIFF
--- a/cui-client/main.c
+++ b/cui-client/main.c
@@ -37,7 +37,7 @@ char_is_alphanumeric(char c)
 }
 
 static void
-cursor_erase_line()
+cursor_erase_line(void)
 {
   printf("\x1b[2K");  // clear line
   printf("\x1b[0G");  // move cursor to the head of the line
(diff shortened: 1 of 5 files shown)


watasuke102 (approved):

shierote (approved):
[nice to have]
Can you detect that by compiler or by clang formatter?
(maybe I will forget...)

Aki-7:
It seems that writing `void func();` itself is not bad, but different from `void func(void)`, which means clang-format or compiler cannot say this `void func()` is good or not.

Issue #89, opened 2022-07-28
Pointer


Aki-7:
- [x] show cursor at the center of arbitrary screen.
- [x] move the cursor according to the event from input device.
  - be careful that the event from input devices will be also used by 3D space, which means we will have to switch the handler of the events according to the current display system type.
  - ignore boundaries of screens, jumping to other screens.
- [x] take account of boundaries of screens, jumping between screens.
- [x] send pointer events to clients. 
  - we need to define the location of views before.

watasuke102:
All tasks are finally implemented

PR #95, opened 2022-08-01
add zn_cursor

watasuke102:
## Context

implement a part of #89 

## Summary

The cursor is just shown on center of the screen. Cannot move.

## How to check behavior

1. start zen
2. appear the cursor on center of the screen

### DIFF
--- a/include/zen/cursor.h
+++ b/include/zen/cursor.h
@@ -1,0 +1,27 @@
+#ifndef ZEN_CURSOR_H
+#define ZEN_CURSOR_H
+
+#include <wayland-server.h>
+#include <wlr/render/wlr_texture.h>
+#include <wlr/types/wlr_xcursor_manager.h>
+
+#include "zen/input-device.h"
+#include "zen/scene/screen.h"
+#include "zen/seat.h"
+
+struct zn_cursor {
+  int x, y;
+  struct zn_screen* screen;
+  struct wlr_texture* texture;
+  struct wlr_xcursor_manager* xcursor_manager;
+
+  struct wl_listener add_screen_signal;
+};
+
+void zn_cursor_set_image(struct zn_cursor* self, const char* name);
+
+struct zn_cursor* zn_cursor_create(void);
+
+void zn_cursor_destroy(struct zn_cursor* self);
+
+#endif  // ZEN_CURSOR_H
--- a/include/zen/cursor.h
+++ b/include/zen/cursor.h
@@ -15,8 +15,8 @@ struct zn_cursor {
   struct wlr_texture* texture;
   struct wlr_xcursor_manager* xcursor_manager;
 
-  struct wl_listener new_screen_signal;
-  struct wl_listener destroy_screen_signal;
+  struct wl_listener new_screen_listener;
+  struct wl_listener destroy_screen_listener;
 };
 
 void zn_cursor_render(struct zn_cursor* self, struct zn_screen* screen,
--- a/include/zen/cursor.h
+++ b/include/zen/cursor.h
@@ -15,7 +15,7 @@ struct zn_cursor {
   struct wlr_texture* texture;
   struct wlr_xcursor_manager* xcursor_manager;
 
-  struct wl_listener add_screen_signal;
+  struct wl_listener new_screen_signal;
 };
 
 struct zn_cursor* zn_cursor_create(void);
--- a/include/zen/cursor.h
+++ b/include/zen/cursor.h
@@ -11,12 +11,15 @@
 
 struct zn_cursor {
   int x, y;
-  struct zn_screen* screen; // nullable
+  struct zn_screen* screen;  // nullable
   struct wlr_texture* texture;
   struct wlr_xcursor_manager* xcursor_manager;
 
   struct wl_listener new_screen_signal;
 };
+
+void zn_cursor_render(struct zn_cursor* self, struct zn_screen* screen,
+    struct wlr_renderer* renderer);
 
 struct zn_cursor* zn_cursor_create(void);
 
--- a/include/zen/cursor.h
+++ b/include/zen/cursor.h
@@ -18,8 +18,6 @@ struct zn_cursor {
   struct wl_listener add_screen_signal;
 };
 
-void zn_cursor_set_image(struct zn_cursor* self, const char* name);
-
 struct zn_cursor* zn_cursor_create(void);
 
 void zn_cursor_destroy(struct zn_cursor* self);
--- a/include/zen/cursor.h
+++ b/include/zen/cursor.h
@@ -19,9 +19,6 @@ struct zn_cursor {
   struct wl_listener destroy_screen_listener;
 };
 
-void zn_cursor_render(struct zn_cursor* self, struct zn_screen* screen,
-    struct wlr_renderer* renderer);
-
 struct zn_cursor* zn_cursor_create(void);
 
 void zn_cursor_destroy(struct zn_cursor* self);
--- a/include/zen/cursor.h
+++ b/include/zen/cursor.h
@@ -16,6 +16,7 @@ struct zn_cursor {
   struct wlr_xcursor_manager* xcursor_manager;
 
   struct wl_listener new_screen_signal;
+  struct wl_listener destroy_screen_signal;
 };
 
 void zn_cursor_render(struct zn_cursor* self, struct zn_screen* screen,
--- a/include/zen/cursor.h
+++ b/include/zen/cursor.h
@@ -11,7 +11,7 @@
 
 struct zn_cursor {
   int x, y;
-  struct zn_screen* screen;
+  struct zn_screen* screen; // nullable
   struct wlr_texture* texture;
   struct wlr_xcursor_manager* xcursor_manager;
 
--- a/include/zen/scene/screen-layout.h
+++ b/include/zen/scene/screen-layout.h
@@ -10,7 +10,7 @@ struct zn_screen_layout {
   struct wl_list screens;  // zn_screen::link
 
   struct {
-    // Emited when a zn_screen is added
+    // Emitted when a zn_screen is added
     struct wl_signal new_screen;  // (struct zn_screen *)
   } events;
 };
--- a/include/zen/scene/screen-layout.h
+++ b/include/zen/scene/screen-layout.h
@@ -12,8 +12,6 @@ struct zn_screen_layout {
   struct {
     // this is emited when zn_screen is added to zn_screen_layout.
     struct wl_signal new_screen;  // (struct zn_screen *)
-    // this is emited when zn_screen is destroyed.
-    struct wl_signal destroy_screen; // (struct zn_screen *)
   } events;
 };
 
--- a/include/zen/scene/screen-layout.h
+++ b/include/zen/scene/screen-layout.h
@@ -10,7 +10,7 @@ struct zn_screen_layout {
   struct wl_list screens;  // zn_screen::link
 
   struct {
-    // this is emited when zn_screen is added to zn_screen_layout.
+    // Emited when a zn_screen is added
     struct wl_signal new_screen;  // (struct zn_screen *)
   } events;
 };
--- a/include/zen/scene/screen-layout.h
+++ b/include/zen/scene/screen-layout.h
@@ -8,6 +8,10 @@ struct zn_screen;
 
 struct zn_screen_layout {
   struct wl_list screens;  // zn_screen::link
+
+  // this is emited when zn_screen added to zn_screen_layout.
+  struct wl_signal add_screen;  // (struct zn_screen *)
+  // TODO: remove_screen signal is required?
 };
 
 void zn_screen_layout_add(
--- a/include/zen/scene/screen-layout.h
+++ b/include/zen/scene/screen-layout.h
@@ -9,9 +9,12 @@ struct zn_screen;
 struct zn_screen_layout {
   struct wl_list screens;  // zn_screen::link
 
-  // this is emited when zn_screen added to zn_screen_layout.
-  struct wl_signal add_screen;  // (struct zn_screen *)
-  // TODO: remove_screen signal is required?
+  struct {
+    // this is emited when zn_screen is added to zn_screen_layout.
+    struct wl_signal new_screen;  // (struct zn_screen *)
+    // this is emited when zn_screen is destroyed.
+    struct wl_signal destroy_screen; // (struct zn_screen *)
+  } events;
 };
 
 void zn_screen_layout_add(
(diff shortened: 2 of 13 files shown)


zen/cursor.c
```diff
@@ -0,0 +1,81 @@
...
+    zn_error("Failed to create wlr_xcursor");
+    goto err_free;
+  }
+  wlr_xcursor_manager_load(self->xcursor_manager, 1.f);
+
+  self->screen = NULL;
+  zn_cursor_set_image(self, "left_ptr");
+
+  struct zn_server* server = zn_server_get_singleton();
+  struct zn_screen_layout* screen_layout = server->scene->screen_layout;
```
Aki-7, line 61: We put all the declaration at the beginning of the scope

include/zen/cursor.h
```diff
@@ -0,0 +1,27 @@
...
+#include <wlr/render/wlr_texture.h>
+#include <wlr/types/wlr_xcursor_manager.h>
+
+#include "zen/input-device.h"
+#include "zen/scene/screen.h"
+#include "zen/seat.h"
+
+struct zn_cursor {
+  int x, y;
+  struct zn_screen* screen;
```
Aki-7, line 14:
```suggestion
  struct zn_screen* screen; // nullable
```

include/zen/cursor.h
```diff
@@ -0,0 +1,27 @@
...
+  struct wlr_texture* texture;
+  struct wlr_xcursor_manager* xcursor_manager;
+
+  struct wl_listener add_screen_signal;
+};
+
+void zn_cursor_set_image(struct zn_cursor* self, const char* name);
```
Aki-7, line 21: We don't need public access of this function for now.

zen/cursor.c
```diff
@@ -0,0 +1,81 @@
...
+  struct zn_cursor* self = zn_container_of(listener, self, add_screen_signal);
+  if (self->screen == NULL) {
+    self->screen = data;
+    self->x = self->screen->output->wlr_output->width / 2;
+    self->y = self->screen->output->wlr_output->height / 2;
+  }
+}
+
+void
+zn_cursor_set_image(struct zn_cursor* self, const char* name)
```
Aki-7, line 24: This function can fail, so this should return true or false.
zn_cursor_create needs to handle the result.

zen/cursor.c
```diff
@@ -0,0 +1,81 @@
...
+{
+  struct zn_server* server = zn_server_get_singleton();
+  struct wlr_xcursor* xcursor;
+
+  xcursor = wlr_xcursor_manager_get_xcursor(self->xcursor_manager, name, 1.f);
+  if (xcursor == NULL) {
+    zn_error("Failed to get xcursor (%s)", name);
+    return;
+  }
+  struct wlr_xcursor_image* image = xcursor->images[0];
```
Aki-7, line 34: Declaration at the beginning of scope

zen/cursor.c
```diff
@@ -0,0 +1,81 @@
...
+  }
+  wlr_xcursor_manager_load(self->xcursor_manager, 1.f);
+
+  self->screen = NULL;
+  zn_cursor_set_image(self, "left_ptr");
+
+  struct zn_server* server = zn_server_get_singleton();
+  struct zn_screen_layout* screen_layout = server->scene->screen_layout;
+  self->add_screen_signal.notify = zn_cursor_handle_add_screen;
+  wl_signal_add(&screen_layout->add_screen, &self->add_screen_signal);
```
Aki-7, line 63: We should also handle the case where cursor->screen is destroyed.

zen/render-2d.c
```diff
@@ -26,6 +27,19 @@ void
...
+      0, 0, 1   //
+  };
+
   wl_list_for_each(view, &screen->views, link)
       zn_render_2d_view(view, renderer);
+
+  if (cursor->screen == screen && cursor->texture) {
+    wlr_render_texture(
+        renderer, cursor->texture, transform, cursor->x, cursor->y, 1.f);
+  }
```
Aki-7, line 44: I wanna new function that render the cursor.
I wanna keep this function (zn_render_2d_screen) as simple as possible, so that developers can easily grasp the entire rendering process.

`transform` matrix you newly declared is only for cursor and that kind of things should not appear in this function.

include/zen/scene/screen-layout.h
```diff
@@ -8,6 +8,10 @@ struct zn_screen;
 
 struct zn_screen_layout {
   struct wl_list screens;  // zn_screen::link
+
+  // this is emited when zn_screen added to zn_screen_layout.
+  struct wl_signal add_screen;  // (struct zn_screen *)
+  // TODO: remove_screen signal is required?
```
Aki-7, line 14: Yes, in this PR.

include/zen/scene/screen-layout.h
```diff
@@ -8,6 +8,10 @@ struct zn_screen;
 
 struct zn_screen_layout {
   struct wl_list screens;  // zn_screen::link
+
+  // this is emited when zn_screen added to zn_screen_layout.
+  struct wl_signal add_screen;  // (struct zn_screen *)
```
Aki-7, line 13:
```suggestion
  struct {
    new_screen; // (struct zn_screen *)
  } events;
```

Following the convention of wlroots, I feel that signals should be placed in events struct.
(I wanna fix `zn_display_system::switch_signal`.)

Aki-7 (changes requested):
Lovely small cursor!

zen/render-2d.c
watasuke102: It will be implemented as `zn_cursor_render()`, right? In this case, should it compare cursor->screen and screen in that function? (I mean, I'm concerning whether `if` statement move to that function or not)

zen/cursor.c
Aki-7: Do you mean cursor->screen?

Aki-7: In the case where cursor->screen is equal to destroyed screen, which means a physical monitor where the cursor is shown get disconnected, it's natural to move the cursor to another screen(monitor) if exists.
If no other screen exists, which means no physical monitor is connected, it's natural that cursor->screen is null

zen/render-2d.c
Aki-7: Which one is ok, but my preference is that checking if the screen has cursor or not is done in the early part of `zn_cursor_render`, which makes `zn_render_2d_screen` more simple and is not so much confusing.

zen/cursor.c
```diff
@@ -0,0 +1,120 @@
...
+
+  if (self->screen != data) {
+    return;
+  }
+
+  if (wl_list_empty(&screen_layout->screens)) {
+    self->screen = NULL;
+  } else {
+    self->screen =
+        zn_container_of(&screen_layout->screens.next, self->screen, link);
```
watasuke102, line 39: It seems all of them are fixed. I want to check, especially here.

zen/cursor.c
```diff
@@ -0,0 +1,120 @@
...
+  free(self);
+
+err:
+  return NULL;
+}
+
+void
+zn_cursor_destroy(struct zn_cursor* self)
+{
+  wl_list_remove(&self->new_screen_signal.link);
```
Aki-7, line 116:
```suggestion
  wl_list_remove(&self->new_screen_signal.link);
  wl_list_remove(&self->destroy_screen_signal.link);
```

include/zen/cursor.h
```diff
@@ -0,0 +1,29 @@
...
+#include "zen/scene/screen.h"
+#include "zen/seat.h"
+
+struct zn_cursor {
+  int x, y;
+  struct zn_screen* screen;  // nullable
+  struct wlr_texture* texture;
+  struct wlr_xcursor_manager* xcursor_manager;
+
+  struct wl_listener new_screen_signal;
```
Aki-7, line 18:
```suggestion
  struct wl_listener new_screen_listener;
```

zen/cursor.c
```diff
@@ -0,0 +1,120 @@
...
+      image->width * 4, image->width, image->height, image->buffer);
+
+  self->screen = NULL;
+
+  self->new_screen_signal.notify = zn_cursor_handle_new_screen;
+  wl_signal_add(&screen_layout->events.new_screen, &self->new_screen_signal);
+
+  self->destroy_screen_signal.notify = zn_cursor_handle_destroy_screen;
+  wl_signal_add(
+      &screen_layout->events.destroy_screen, &self->destroy_screen_signal);
```
Aki-7, line 99: It's a little bit strange for me to create a signal at `screen_layout` that notifies the destruction of `screen`.
The signal that notifies screen destruction seems natural to be placed on `screen`.

You can create `zn_cursor::screen_destroy_listener` and set `zn_cursor::screen_destroy_listener.notify` in the constructor of `zn_cursor`, but do not register the listener to any signals.

Instead, when `zn_cursor_handle_new_screen` is called, you can add the listener to the screen's destroy signal.

zen/cursor.c
Aki-7: In addition to this, please update the destroy order in `zn_server_destroy`
from
```c
void
{
  if (self->xwayland) wlr_xwayland_destroy(self->xwayland);
  zn_input_manager_destroy(self->input_manager);
  zn_display_system_destroy(self->display_system);
  free(self->socket);
  zn_scene_destroy(self->scene); // position of this
  wlr_allocator_destroy(self->allocator);
  wlr_renderer_destroy(self->renderer);
  wlr_backend_destroy(self->backend);
  server_singleton = NULL;
  free(self);
}
```
to
```c
void
{
  if (self->xwayland) wlr_xwayland_destroy(self->xwayland);
  zn_input_manager_destroy(self->input_manager);
  zn_display_system_destroy(self->display_system);
  free(self->socket);
  wlr_allocator_destroy(self->allocator);
  wlr_renderer_destroy(self->renderer);
  wlr_backend_destroy(self->backend);
  zn_scene_destroy(self->scene);
  server_singleton = NULL;
  free(self);
}
```

Your current code sometimes exit with segmentation fault during termination.

Aki-7 (changes requested):

zen/cursor.c
```diff
@@ -0,0 +1,107 @@
...
+
+static void
+zn_cursor_handle_destroy_screen(struct wl_listener* listener, void* data)
+{
+  UNUSED(data);
+  struct zn_cursor* self = zn_container_of(listener, self, new_screen_listener);
+  struct zn_server* server = zn_server_get_singleton();
+  struct zn_screen_layout* screen_layout = server->scene->screen_layout;
+
+  wl_list_remove(&self->destroy_screen_listener.link);
```
Aki-7, line 21:
```suggestion
  wl_list_remove(&self->destroy_screen_listener.link);
  wl_list_init(&self->destroy_screen_listener.link);
```
You need to init the list. Otherwise `wl_list_remove(&self->destroy_screen_listener.link);` is executed again in the destructor, and it's not safe.

zen/cursor.c
```diff
@@ -0,0 +1,107 @@
...
+{
+  struct zn_cursor* self = zn_container_of(listener, self, new_screen_listener);
+  struct zn_screen* screen = data;
+
+  if (self->screen == NULL) {
+    self->screen = screen;
+    self->x = self->screen->output->wlr_output->width / 2;
+    self->y = self->screen->output->wlr_output->height / 2;
+
+    self->destroy_screen_listener.notify = zn_cursor_handle_destroy_screen;
```
Aki-7, line 42: It's enough to do this once in the constructor.

Aki-7 (changes requested):
It's been a long time coming, but we'll soldier on!

zen/cursor.c
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  self->screen = screen;
+}
+
+static void
+zn_cursor_handle_destroy_screen(struct wl_listener* listener, void* data)
+{
+  UNUSED(data);
+  struct zn_cursor* self = zn_container_of(listener, self, new_screen_listener);
```
Aki-7, line 37:
```suggestion
  struct zn_cursor* self = zn_container_of(listener, self, destroy_screen_listener);
```

include/zen/scene/screen-layout.h
```diff
@@ -8,6 +8,11 @@ struct zn_screen;
 
 struct zn_screen_layout {
   struct wl_list screens;  // zn_screen::link
+
+  struct {
+    // Emited when a zn_screen is added
```
Aki-7, line 13:
```suggestion
    // Emitted when a zn_screen is added
```
typo

zen/cursor.c
```diff
@@ -0,0 +1,122 @@
...
+  struct zn_screen_layout* screen_layout = server->scene->screen_layout;
+  struct zn_screen* screen;
+
+  if (wl_list_empty(&screen_layout->screens)) {
+    screen = NULL;
+  } else {
+    screen = zn_container_of(&screen_layout->screens.next, self->screen, link);
+  }
+
+  zn_cursor_set_screen(self, screen);
```
Aki-7, line 48: As the `screen` is not removed form `screen_layout::screens` at this point, searching new screen in this way doesn't work. Try to disconnect your monitor, and you will get error.

So in this case, we should have a signal notifying that a `zn_screen` has removed from a `zn_screen_layout`.
That's so similar to what you have implemented before and I'm sorry that I couldn't properly understand the situation at that point. But I want you to create such signal again with the name of something like `zn_screen_layout::events::screen_removed`, and listener `zn_cursor::screen_removed_listener`.

Aki-7 (changes requested):

zen/cursor.c
watasuke102: I think following code works too. Can I implement like this?

```
  wl_list_for_each(screen, &screen_layout->screens, link)
  {
    if (screen != self->screen) {
      zn_cursor_set_screen(self, screen);
      return;
    }
  }

  zn_cursor_set_screen(self, NULL);
```

Aki-7: I actually implemented the exact same way locally to check behavior 10 min ago, but that code is not extendable. I mean it is difficult to add other process later.

So at first, I tried to suggest like this,
```c
  bool find = false;

  wl_list_for_each(screen, &screen_layout->screens, link)
  {
    if (screen != self->screen) {
      find = true;
      break;
    }
  }

  zn_cursor_set_screen(self, find ? screen : NULL);
```

and I know it's much easier to implement from now on.
However, we should be careful not to forget to seek the best implementation.
I personally think that implementing like my suggestion is better because it's more simple and less confusing.
Not confusing about the `screen` is in the list or not.

Aki-7: I know this PR getting long and exhausting, and I'm partially responsible for this. But let's do our best.

Aki-7 (approved):
LGTM!
Good Job!

watasuke102:
Thank you for review patiently~

Aki-7:
### How to Check behavior

1. build
2. run zen-desktop with one monitor.
3. add another monitor
4. disconnect the monitor where a cursor is shown

The cursor appears in the remaining monitor.

PR #97, opened 2022-08-02
add -Wstrict-prototypes

Aki-7:
## Context

See https://github.com/zigen-project/zen/pull/94#pullrequestreview-1057041890

## Summary

Add compiler parameter that forbidden function declaration/definition with empty parentheses.

**bad (compile error)**
```
void func();
```
**good**
```
void func(void);
```

## How to check behavior

Compile with some empty parentheses function, and you will get compile error.

### DIFF
--- a/include/zen/server.h
+++ b/include/zen/server.h
@@ -44,7 +44,7 @@ struct zn_server {
   int exit_code;
 };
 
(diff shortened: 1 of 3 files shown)


watasuke102 (approved):

shierote (approved):
LGTM!

PR #98, opened 2022-08-04
Pointer

watasuke102:
## Context

implement a part of #89 

## Summary

The cursor is moved corresponding with mouse input. For now, it ignores boundaries of screens and doesn’t jump to other screens.

## How to check behavior

1. start zen
2. move your mouse
3. the cursor is moved

### DIFF
--- a/include/zen/pointer.h
+++ b/include/zen/pointer.h
@@ -5,8 +5,6 @@
 #include <wlr/interfaces/wlr_pointer.h>
 
 struct zn_pointer {
-  struct wlr_pointer* wlr_pointer;
-
   struct wl_listener motion_listener;
 };
 
--- a/include/zen/pointer.h
+++ b/include/zen/pointer.h
@@ -6,6 +6,8 @@
 
 struct zn_pointer {
   struct wlr_pointer* wlr_pointer;
+
+  struct wl_listener motion_relative_listener;
 };
 
 struct zn_pointer* zn_pointer_create(struct wlr_input_device* input_device);
--- a/include/zen/pointer.h
+++ b/include/zen/pointer.h
@@ -7,7 +7,7 @@
 struct zn_pointer {
   struct wlr_pointer* wlr_pointer;
 
-  struct wl_listener motion_relative_listener;
+  struct wl_listener motion_listener;
 };
 
 struct zn_pointer* zn_pointer_create(struct wlr_input_device* input_device);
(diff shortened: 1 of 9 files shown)


zen/cursor.c
```diff
@@ -62,6 +62,13 @@ zn_cursor_handle_new_screen(struct wl_listener* listener, void* data)
   }
 }
 
+void
+zn_cursor_move_relative(struct zn_cursor* self, int dx, int dy)
```
liwii, line 66: Why is it named "relative"? It seems that on wlroot non-relative event is called with "absolute" but the relative one is just called "wlr_pointer_motion_event" (sorry if I'm pointing to a wrong line)

watasuke102: Yes, at wlroots, it called `motion`, and non-relative event is called `absolute`. But I think giving contrasting name is good. (It is my preference, so if you have some reason to follow naming by wlroot, I will make change this)

liwii: Sounds good to me. I would say that if we won't use the "absolute" motion in the future it's reasonable to name it just `zgn_cursor_motion`. But otherwise, it is very good to show the contrast with naming.

watasuke102: That's true.

include/zen/pointer.h
```diff
@@ -0,0 +1,17 @@
+#ifndef ZEN_POINTER_H
+#define ZEN_POINTER_H
+
+#include <wayland-server.h>
+#include <wlr/interfaces/wlr_pointer.h>
+
+struct zn_pointer {
+  struct wlr_pointer* wlr_pointer;
```
Aki-7, line 8: What is this member for? If you are using this member in the next pull request, it's ok to keep this.

zen/pointer.c
```diff
@@ -0,0 +1,55 @@
...
+struct zn_pointer*
+zn_pointer_create(struct wlr_input_device* wlr_input_device)
+{
+  if (!zn_assert(wlr_input_device->type == WLR_INPUT_DEVICE_POINTER,
+          "Wrong type - expect: %d, actual: %d", WLR_INPUT_DEVICE_POINTER,
+          wlr_input_device->type)) {
+    goto err;
+  }
+
+  struct zn_pointer* self;
```
Aki-7, line 30: I want this declaration to be placed at the beginning of the scope. I missed the same code in keyboard.c, please fix it as well.

Aki-7 (changes requested):

Aki-7 (approved):
LGTM!

Issue #82, opened 2022-07-26
wlr_xwayland_create should be called before executing startup command


Aki-7:
## Context

Startup command is the shell command that is executed when the compositor starts.
We can specify the command with -s option

## Problem

Now, the following command fails because xeyes cannot connect to X server (xwayland).

```
zen-desktop -s xeyes
```

## Cause

startup command is executed before initializing xwayland.
e.i., launch_startup_command in main.c should be called after wlr_xwayland_create in server.c

Aki-7:
Thank you @shierote !

PR #99, opened 2022-08-04
Change the timing of creating and starting xwayland with lazy option

shierote:
## Context

fix https://github.com/zigen-project/zen/issues/82.

## Summary

- To use x client in startup command (or `-s` option), xwayland is started with lazy option in `zn_server_create` instead of `zn_server_run`.
- Some corrections, including mistakes that destructor is not written.

## How to check behavior

```
$ zen-desktop -s xeyes
```
Then, the monitor shows xeyes's window.

Without this pr, this command terminates zen-desktop without showing any windows.

### DIFF
--- a/zen/server.c
+++ b/zen/server.c
@@ -122,20 +122,6 @@ zn_server_get_singleton(void)
 int
 zn_server_run(struct zn_server *self)
 {
-  self->xwayland = wlr_xwayland_create(self->display, self->w_compositor, true);
-  if (self->xwayland == NULL) {
-    zn_error("Failed to create xwayland");
-    return EXIT_FAILURE;
-  }
-
-  self->xwayland_new_surface_listener.notify =
-      zn_server_xwayland_new_surface_handler;
-  wl_signal_add(&self->xwayland->events.new_surface,
-      &self->xwayland_new_surface_listener);
-
-  setenv("DISPLAY", self->xwayland->display_name, true);
-  zn_debug("DISPLAY=%s", self->xwayland->display_name);
-
   if (!wlr_backend_start(self->backend)) {
     zn_error("Failed to start backend");
     return EXIT_FAILURE;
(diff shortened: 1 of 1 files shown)


zen/server.c
```diff
@@ -241,16 +227,25 @@ zn_server_create(struct wl_display *display)
...
   if (self->input_manager == NULL) {
     zn_error("Failed to create input manager");
     goto err_socket;
   }
 
-  setenv("WAYLAND_DISPLAY", self->socket, true);
-  xdg = getenv("XDG_RUNTIME_DIR");
-  zn_debug("WAYLAND_DISPLAY=%s", self->socket);
-  zn_debug("XDG_RUNTIME_DIR=%s", xdg);
+  self->xwayland = wlr_xwayland_create(self->display, self->w_compositor, true);
```
liwii, line 241: Could we move L241 ~ 248 before the `zn_input_manager_create`? It seems a bit weird that input-related code is divided by this xwayland block.

liwii (changes requested):

zen/server.c
shierote: Thanks for your review!
Do you recognize `zn_input_manager_create`  and `self->new_input_listener.notify = zn_server_new_input_handler;..` as input-related code? (confirmation)

I intended that xwayland would be optional in the future, so I am writing it at the bottom of the member creation step.
However, this intention is not a strong one, so if you feel that it is wired after hearing this intention, then I will fix it as you said.

Aki-7: That doesn't seem weird for me. Rather than the xwayland block being placed between the input manager-related code, it seems more natural to think that it was placed at the end of the member creation block and before the register event listener block.

zen/server.c
```diff
@@ -301,9 +304,10 @@ zn_server_create(struct wl_display *display)
 void
 zn_server_destroy(struct zn_server *self)
 {
-  if (self->xwayland) wlr_xwayland_destroy(self->xwayland);
+  wlr_xwayland_destroy(self->xwayland);
   zn_input_manager_destroy(self->input_manager);
   zn_display_system_destroy(self->display_system);
+  zn_immersive_backend_destroy(self->immersive_backend);
```
Aki-7, line 310: thx

Aki-7 (approved):

zen/server.c
watasuke102: Hmm, it is actually true that that block is placed between member creation block and before the register event listener block. But if divide member creation block more, I think it is also true that input-related code is divided by xwayland block.

Pointed out lines seem to relate xwayland, so unless there is no reason to put these above creating input-related code, it seems moving this process is good.

Aki-7: Rather xwayland block beeing between member creation block and event registration block, it seems one of the member creation block for me. If someone want to add another member creation block that is related to input manager, it's ok to place the block between xwayland block and input manager block.

PR #105, opened 2022-08-14
Send pointer event

watasuke102:
## Context

implement a part of #89 

## Summary

Give position information to each views, and send enter/leave/motion/axis event from pointer to client.

## How to check behavior

1. start zen with -s option
    - `weston-clickdot` is recommended for testing `motion` event
    - `weston-cliptest` is recommended for testing `axis` event
3. move the cursor or scroll on view
4. events are sended

### DIFF
--- a/include/zen/scene/view.h
+++ b/include/zen/scene/view.h
@@ -28,7 +28,7 @@ struct zn_view {
   struct wl_list link;  // zn_screen::views;
 };
 
-void zn_view_get_box(struct zn_view *self, struct wlr_box *box);
+void zn_view_get_fbox(struct zn_view *self, struct wlr_fbox *fbox);
 
 bool zn_view_is_mapped(struct zn_view *self);
 
--- a/include/zen/scene/view.h
+++ b/include/zen/scene/view.h
@@ -32,6 +32,8 @@ void zn_view_get_box(struct zn_view *self, struct wlr_box *box);
 
 bool zn_view_is_mapped(struct zn_view *self);
 
+void zn_view_focus(struct zn_view *self);
+
 void zn_view_map_to_screen(struct zn_view *self, struct zn_screen *screen);
 
 void zn_view_unmap(struct zn_view *self);
--- a/include/zen/scene/view.h
+++ b/include/zen/scene/view.h
@@ -19,12 +19,16 @@ enum zn_view_type {
 };
 
 struct zn_view {
+  int x, y;
+
   enum zn_view_type type;
 
   const struct zn_view_impl *impl;
 
   struct wl_list link;  // zn_screen::views;
 };
+
+void zn_view_get_box(struct zn_view *self, struct wlr_box *box);
 
 bool zn_view_is_mapped(struct zn_view *self);
 
--- a/include/zen/scene/view.h
+++ b/include/zen/scene/view.h
@@ -20,7 +20,6 @@ enum zn_view_type {
 
 struct zn_view {
   int x, y;
-  bool is_moving;
 
   enum zn_view_type type;
 
@@ -32,8 +31,6 @@ struct zn_view {
 void zn_view_get_box(struct zn_view *self, struct wlr_box *box);
 
 bool zn_view_is_mapped(struct zn_view *self);
-
-void zn_view_move(struct zn_view *self, int dx, int dy);
 
 void zn_view_focus(struct zn_view *self);
 
--- a/include/zen/scene/view.h
+++ b/include/zen/scene/view.h
@@ -19,7 +19,7 @@ enum zn_view_type {
 };
 
 struct zn_view {
-  int x, y;
+  double x, y;
 
   enum zn_view_type type;
 
--- a/include/zen/scene/view.h
+++ b/include/zen/scene/view.h
@@ -20,6 +20,7 @@ enum zn_view_type {
 
 struct zn_view {
   int x, y;
+  bool is_moving;
 
   enum zn_view_type type;
 
@@ -31,6 +32,8 @@ struct zn_view {
 void zn_view_get_box(struct zn_view *self, struct wlr_box *box);
 
 bool zn_view_is_mapped(struct zn_view *self);
+
+void zn_view_move(struct zn_view *self, int dx, int dy);
 
 void zn_view_focus(struct zn_view *self);
 
--- a/include/zen/scene/view.h
+++ b/include/zen/scene/view.h
@@ -34,6 +34,8 @@ bool zn_view_is_mapped(struct zn_view *self);
 
 void zn_view_focus(struct zn_view *self);
 
+void zn_view_unfocus(struct zn_view *self);
+
 void zn_view_map_to_screen(struct zn_view *self, struct zn_screen *screen);
 
 void zn_view_unmap(struct zn_view *self);
--- a/include/zen/scene/view.h
+++ b/include/zen/scene/view.h
@@ -32,8 +32,6 @@ void zn_view_get_box(struct zn_view *self, struct wlr_box *box);
 
 bool zn_view_is_mapped(struct zn_view *self);
 
-void zn_view_focus(struct zn_view *self);
-
 void zn_view_map_to_screen(struct zn_view *self, struct zn_screen *screen);
 
 void zn_view_unmap(struct zn_view *self);
--- a/include/zen/scene/view.h
+++ b/include/zen/scene/view.h
@@ -34,8 +34,6 @@ bool zn_view_is_mapped(struct zn_view *self);
 
 void zn_view_focus(struct zn_view *self);
 
-void zn_view_unfocus(struct zn_view *self);
-
 void zn_view_map_to_screen(struct zn_view *self, struct zn_screen *screen);
 
 void zn_view_unmap(struct zn_view *self);
--- a/zen/pointer.c
+++ b/zen/pointer.c
@@ -93,12 +93,12 @@ zn_pointer_create(struct wlr_input_device* wlr_input_device)
   wl_signal_add(
       &wlr_input_device->pointer->events.motion, &self->motion_listener);
 
-  self->axis_listener.notify = zn_pointer_handle_axis;
-  wl_signal_add(&wlr_input_device->pointer->events.axis, &self->axis_listener);
-
   self->button_listener.notify = zn_pointer_handle_button;
   wl_signal_add(
       &wlr_input_device->pointer->events.button, &self->button_listener);
+
+  self->axis_listener.notify = zn_pointer_handle_axis;
+  wl_signal_add(&wlr_input_device->pointer->events.axis, &self->axis_listener);
 
   self->frame_listener.notify = zn_pointer_handle_frame;
   wl_signal_add(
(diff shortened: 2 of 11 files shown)


zen/scene/screen.c
```diff
@@ -2,6 +2,25 @@
...
+  {
+    zn_view_get_box(view, &box);
+
+    if (wlr_box_contains_point(&box, x, y)) {
+      return view;
+    }
+  }
+
+  return NULL;
+}
```
Aki-7, line 23: This code is good, but could you please render the views in reverse order (in render-2d.c, zn_render_2d_screen)?
I think screen->views are ordered from from to bottom.
We need to render views from bottom to top.

I know it's a bug made before this PR, but please.

include/zen/scene/view.h
```diff
@@ -19,15 +19,21 @@ enum zn_view_type {
 };
 
 struct zn_view {
+  int x, y;
```
Aki-7, line 22:
```suggestion
  double x, y;
```

zen/scene/view.c
```diff
@@ -11,9 +25,42 @@ zn_view_is_mapped(struct zn_view *self)
...
+      break;
+    case ZN_VIEW_XWAYLAND:
+      xwayland_view = zn_container_of(self, xwayland_view, base);
+      if (xwayland_view->wlr_xwayland_surface->surface != current_surface) {
+        zn_xwayland_view_focus(xwayland_view);
+      }
+      break;
+  }
+}
+
```
Aki-7, line 53: 1. Rather than createing zn_xdg_toplevel_view_focus and zn_xwayland_view_focus, it's better to add new function pointer member to zn_view_impl, by which we can make xdg_toplevel and xwayland views more uniform.
2. Most logic regarding views should be described in view.c rather than xdg-toplevel-view.c or xwayland-view.c, and only view type (xdg or xwayland) dependent processes should be delegated to them through zn_view_impl interface. That is, the name of zn_xwayland_view_set_activated is better than zn_xwayland_view_focus.
3. Would you explain what is wlr_xdg_toplevel_set_activate, wlr_xwayland_surface_activate and wlr_xwayland_surface_stack? Are those necessary now?

zen/xdg-toplevel-view.c
```diff
@@ -39,6 +28,7 @@ zn_xdg_toplevel_view_map(struct wl_listener* listener, void* data)
   screen = zn_container_of(scene->screen_layout->screens.next, screen, link);
 
   zn_view_map_to_screen(&self->base, screen);
+  zn_xdg_toplevel_view_focus(self);
```
Aki-7, line 31: I think views should be focused every time it's mapped, so it's better to focus a view in zn_view_map_to_screen.

Aki-7: And there is no corresponding code in xwayland-view.c

Aki-7 (changes requested):

zen/scene/view.c
watasuke102: `wlr_(xwayland|xdg_toplevel)_set_activate()` is core of this `focus` process. `wlr_xwayland_surface_restack()` is called here because sway called this at simillar function. But in fact, I don't know how behave this because sending event to xwayland client doesn't work well yet, so I will remove this.

These `focus` process doesn't required in terms of behavior, but only make it easier to visually see which view is focusing. I can remove these process too, but are these **really** unnecessary?

Aki-7: > wlr_(xwayland|xdg_toplevel)_set_activate() is core of this focus process.

What are these functions doing?

> sending event to xwayland client doesn't work well yet.

What do you mean? What events? Leaving those event not working doesn't make problems?

> These focus process doesn't required in terms of behavior, but only make it easier to visually see which view is focusing.

Does 'visually' mean the code is visually understandable for programmers?

watasuke102: > What are these functions doing?

Sorry for these remark because I'm lack of complehension, but May I ask what are you intending?

> What events?

It is pointer events such as motion.

> Leaving those event not working doesn't make problems?

No, so I will deal this problem later (in this PR).

> Does 'visually' mean the code is visually understandable for programmers?

No, it means, focused view's border (including the top of the view, usually contains title text and minimize/maximize/close button) is shown more brightly than not focused view's border (try to add `return;` line to the first line of `zn_xdg_toplevel_view_focus()`). As I mentioned before comment, these `focus` process doesn't required (in terms of behavior).

zen/scene/screen.c
watasuke102: Is following code resolves this? (441bc46)

```diff
- wl_list_for_each(view, &self->views, link)
+ wl_list_for_each_reverse(view, &self->views, link)
```

zen/scene/view.c
Aki-7: > Sorry for these remark because I'm lack of complehension, but May I ask what are you intending?

It's just a question. I want to know what the `set_activated` means. Sending set_activated event to clients? Moving the view forward? (I think neither of them). I want to know what `set_activated` functions do, and also I want you to know it.

> No, it means, ...

Do you mean that kind of visual cue for focusing view has already been realized in the current code (https://github.com/zigen-project/zen/pull/105/commits/84f1de52b927df52532d6110fdd8ca956bdeb3c5)?

I couldn't confirm that a view's frame got more bright when the view is focused.

zen/scene/screen.c
Aki-7: I don't think so. 441bc46 assumes screen::views has views from back to front. It's ok, but then a view should be inserted at the last of screen::views when created.

zen/scene/view.c
watasuke102: > I would like you to find it out.

Needless to say I implemented this because I think it is necessary, but I also understand that it may be my prefer rather than wide perspective, so I requested review.

You asked at first to me, `Are those necessary now?`, I interpreted this as you think it is not necessary now by some reasons, but I couldn't guess the reasons. May I ask about its reason?

> Sending set_activated event to clients? Moving the view forward? (I think neither of them)

Ok, it really helpful because I didn't know how much detail to describe.

`wlr_xdg_toplevel_set_activated(wlr_xdg_surface* surface, bool activated)` is set `activated` to `surface->scheduled.activated`, and add surface configurating event (`surface_send_configure()`) to event loop.  
On configurating, if the surface is xdg_toplevel and `surface->scheduled.activated` is true, add activated status to xdg_toplevel's state.

I'm doesn't x11 well, so I couldn't understand  `wlr_xwayland_surface_activate(wlr_xwayland_surface *xsurface, bool activated)`, but I found that this function finally call `xcb_change_property()`. It looks updating active window... (understanding `xcb_atom_t` is might helpful, if required more understanding, I learn about this)

> I couldn't confirm that a view's frame got more bright when the view is focused

Sorry, on the current code, I forget to unfocus view (that's why I said that "add `return;` to..."), so differences are not shown.

Aki-7: > You asked at first to me, Are those necessary now?, I interpreted this as you think it is not necessary now by some reasons, but I couldn't guess the reasons. May I ask about its reason?

No, it was basically just a question but I was also thinking that if there was no positive answer it would be unnecessary.

Now you mean the current `zn_view_focus` func sends [xdg_toplevel::configure](https://wayland.app/protocols/xdg-shell#xdg_toplevel:event:configure) events to clients with active state included in its `states` arg, and then clients update window theme (frame, etc.) to be shown as active. That's the reason you implement `zn_(xdg_toplevel|xwayland)_view_focus`, right?

I didn't understand the meaning of `states` argument of xdg_toplevel::configure events, so this explanation is helpful for me. Thanks!

As you mentioned, if you include view activation feature in this PR, I want you to also implement deactivating feature.

watasuke102: > That's the reason you implement zn_(xdg_toplevel|xwayland)_view_focus

That's right. 

Now, I think implementing `unfocus` process is little difficult, and `focus` process can be better by using `impl` (suggested at first). So I will remove `focus` process by this PR, and create another PR a while leter.

zen/pointer.c
```diff
@@ -14,8 +16,51 @@ zn_pointer_handle_motion(struct wl_listener* listener, void* data)
...
+  view = zn_screen_get_view_at(cursor->screen, cursor->x, cursor->y);
+  if (!view) {
+    wlr_seat_pointer_notify_clear_focus(seat);
+    return;
+  }
+
+  surface = view->impl->get_wlr_surface(view);
+  if (surface) {
+    view_x = cursor->x - view->x;
+    view_y = cursor->y - view->y;
```
Aki-7, line 35: Like sway, view_(x|y) should be calculated by zn_screen_get_view_at. The reason is that obtaining view_(x|y) is the responsibility of the scene side, such as view and screen, and should not be calculated in functions that handle pointer motion.

Aki-7 (changes requested):

zen/scene/view.c
```diff
@@ -14,7 +28,15 @@ zn_view_is_mapped(struct zn_view *self)
...
-  wl_list_insert(&screen->views, &self->link);
+  struct wlr_box screen_box;
+  struct wlr_fbox view_fbox;
+  zn_screen_get_box(screen, &screen_box);
+  zn_view_get_fbox(self, &view_fbox);
+
+  self->x = (screen_box.width / 2) - (view_fbox.width / 2);
+  self->y = (screen_box.height / 2) - (view_fbox.height / 2);
+
+  wl_list_insert(screen->views.prev, &self->link);
```
Aki-7, line 39: Sorry for additional review. Please add comment to zn_screen::views.

"List of mapped zn_view in z-order, from bottom to top"

zen/scene/screen.c
```diff
@@ -2,6 +2,33 @@
...
+  wl_list_for_each_reverse(view, &self->views, link)
+  {
+    struct wlr_box box;
+    zn_view_get_fbox(view, &fbox);
+    box.x = fbox.x;
+    box.y = fbox.y;
+    box.width = fbox.width;
+    box.height = fbox.height;
+
+    if (wlr_box_contains_point(&box, x, y)) {
```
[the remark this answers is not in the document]
watasuke102: because `wlr_box_contains_point()` requires wlr_box, not wlr_fbox.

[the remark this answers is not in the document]
Aki-7: I see no reason why `wlr_box_contains_point` should be used.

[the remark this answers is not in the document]
Aki-7: We are using screen world described in double precision floating point. Containment checks should also be performed with double precision floating point.

zen/scene/view.c
Aki-7: I wanted to add a comment to the definition of screen struct, views member in screen.h. Sorry for the misunderstanding.

Aki-7:
new comment: https://github.com/zigen-project/zen/pull/105#discussion_r946399592

You will fix xwayland problem in this PR, right?

watasuke102:
Yes, I'm working about it.

Aki-7 (approved):
LGTM!

Great progress!!!

Issue #103, opened 2022-08-14
Add rule to clang-format


watasuke102:
- Set true: `InsertBraces` rule
- Set: `PointerAlignment` rule
  - "Left" (`int* a`) or "Right" (`int *a`) or "Middle" (`int * a`) ?

watasuke102:
`InsertBraces` rule is too latest? (from clang-format 15, but current version of Ubuntu 22.04's clang-format is 14...)

Aki-7:
We can consider introducing clang-tidy.

watasuke102:
Ok, it seems to work so good.

shierote:
> "Right"

LGTM!

PR #106, opened 2022-08-15
key binding

Aki-7:
## Context

Keyboard shortcut is short-hand way to check GUI shell functionalities we will implement from now on (switching focus, moving between virtual desktop, etc.).

So I implemented keyboard shortcut basis first.

## Summary

Developers can add keyboard shortcut using

```c
struct zn_binding* zn_input_manager_add_key_binding(
    struct zn_input_manager* self, uint32_t key, uint32_t modifiers,
    zn_key_binding_handler_t handler, void* data);
```

`key` and `modifiers` are shortcut bindings to be used.

When `key` is pressed with `modifiers`, `handler` will be called with `data` in its argument.

See `main.c` for example.

## How to check behavior

1. build
2. start `zen-desktop`
3. press `<alt> + q`

zen-desktop will be terminated gracefully.

4. (option) try to create another keyboard shortcut.

## Notice

Termination shortcut is changed to `<alt> + q`

### DIFF
--- a/include/zen/binding.h
+++ b/include/zen/binding.h
@@ -1,0 +1,32 @@
+#ifndef ZEN_BINDING_H
+#define ZEN_BINDING_H
+
+#include <wayland-server.h>
+
+struct zn_input_manager;
+struct zn_keyboard;
+
+typedef void (*zn_key_binding_handler_t)(
+    uint32_t time_msec, uint32_t key, void *data);
+
+struct zn_binding {
+  uint32_t key;
+  uint32_t modifiers;
+  zn_key_binding_handler_t handler;
+  void *data;
+  struct wl_list link;  // zn_input_manager::key_binding_list
+};
+
+/**
+ * @return true if the key event is handled
+ */
+bool zn_binding_notify_key(struct zn_binding *self, uint32_t time_msec,
+    uint32_t key, enum wl_keyboard_key_state state,
+    struct zn_keyboard *keyboard);
+
+struct zn_binding *zn_binding_create(uint32_t key, uint32_t modifiers,
+    zn_key_binding_handler_t handler, void *data);
+
+void zn_binding_destroy(struct zn_binding *self);
+
+#endif  //  ZEN_BINDING_H
--- a/zen/binding.c
+++ b/zen/binding.c
@@ -1,0 +1,57 @@
+#include "zen/binding.h"
+
+#include <zen-common.h>
+
+#include "zen/input-manager.h"
+#include "zen/server.h"
+
+bool
+zn_binding_notify_key(struct zn_binding *self, uint32_t time_msec, uint32_t key,
+    enum wl_keyboard_key_state state, struct zn_keyboard *keyboard)
+{
+  struct wlr_keyboard *wlr_keyboard =
+      keyboard->input_device->wlr_input->keyboard;
+
+  if (state == WL_KEYBOARD_KEY_STATE_RELEASED) {
+    return false;
+  }
+
+  if (self->key == key &&
+      self->modifiers == wlr_keyboard_get_modifiers(wlr_keyboard)) {
+    self->handler(time_msec, key, self->data);
+    return true;
+  }
+
+  return false;
+}
+
+struct zn_binding *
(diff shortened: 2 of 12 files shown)


watasuke102:
At this PR, the `key` event by keyboard isn't sent to client, right?

By the way, terminate key binding is alt + **q**, but only this PR's description is wrong (the code, behavior, README is correct).

Aki-7:
> At this PR, the key event by keyboard isn't sent to client, right?

That's right.

> By the way, terminate key binding is alt + q, but only this PR's description is wrong (the code, behavior, README is correct).

Oh, sorry for that. Updated. Thanks!

watasuke102 (approved):
I tried to add another binding, and it works well.

LGTM!

include/zen/binding.h
```diff
@@ -0,0 +1,32 @@
...
+struct zn_keyboard;
+
+typedef void (*zn_key_binding_handler_t)(
+    uint32_t time_msec, uint32_t key, void *data);
+
+struct zn_binding {
+  uint32_t key;
+  uint32_t modifiers;
+  zn_key_binding_handler_t handler;
+  void *data;
```
shierote, line 16: What is the member `data` used for?

zen/binding.c
```diff
@@ -0,0 +1,57 @@
...
+#include "zen/server.h"
+
+bool
+zn_binding_notify_key(struct zn_binding *self, uint32_t time_msec, uint32_t key,
+    enum wl_keyboard_key_state state, struct zn_keyboard *keyboard)
+{
+  struct wlr_keyboard *wlr_keyboard =
+      keyboard->input_device->wlr_input->keyboard;
+
+  if (state == WL_KEYBOARD_KEY_STATE_RELEASED) {
```
shierote, line 15: [question] Should {} be omitted when {} is a single line of processing?

shierote (approved):
LGTM

zen/main.c
```diff
@@ -170,6 +182,10 @@ main(int argc, char *argv[])
     goto err_signal;
   }
 
+  // Terminate the program with a keyboard event for development convenience.
+  zn_input_manager_add_key_binding(server->input_manager, KEY_Q,
+      WLR_MODIFIER_ALT, zn_terminate_binding_handler, NULL);
```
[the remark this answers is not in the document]
Aki-7: Yes, it's just for development convenience. I think terminating with `alt + shift + q` without confirmation is also too easy for a real use case. I think creating logout button later is good idea.

include/zen/binding.h
Aki-7: To store user defined data.

zen/binding.c
watasuke102: This is my preference, but I think all of them should be given {} every time. This rule makes it convinience to extend the process in if statement.

See: #103 (`InsertBraces`)

Aki-7: I don't know. Now both style are mixed, let's fix them in https://github.com/zigen-project/zen/issues/103.

zen/keyboard.c
```diff
@@ -35,8 +56,27 @@ zn_keyboard_create(struct wlr_input_device* wlr_input_device)
...
+  {
+    struct xkb_context* context = xkb_context_new(XKB_CONTEXT_NO_FLAGS);
+    struct xkb_keymap* keymap =
+        xkb_keymap_new_from_names(context, NULL, XKB_KEYMAP_COMPILE_NO_FLAGS);
+    wlr_keyboard_set_keymap(wlr_input->keyboard, keymap);
+    xkb_keymap_unref(keymap);
+    xkb_context_unref(context);
+  }
+
+  wlr_keyboard_set_repeat_info(wlr_input->keyboard, 25, 600);
```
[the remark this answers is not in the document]
Aki-7: I intentionally did not make it a variable.
- those numbers will not be shared in other place.
- the prototype of this function is fully descriptive and making variables like `KEY_REPEATE_RATE`, `KEY_REPEATE_DELAY` is redundant here.

include/zen/binding.h
Aki-7: same as the second arguments of `wl_signal_emit`.

Issue #100, opened 2022-08-06
The cursor point wrong position


watasuke102:
due to be ignoring hotspot

watasuke102:
close by #117

PR #107, opened 2022-08-17
Set cursor surface

watasuke102:
## Context

In other DE, the cursor's appearance is changed when move cursor on view such as resize cursor. But zen can't.

## Summary

Listening wlr_seat's `request_set_cursor` event.

## How to check behavior

1. start zen with -s option (`weston-dnd` is recommended)
2. move the cursor to corners of view or object in view
3. cursor's appearance will be changed

### DIFF
--- a/zen/cursor.c
+++ b/zen/cursor.c
@@ -95,6 +95,12 @@ zn_cursor_move_relative(struct zn_cursor* self, int dx, int dy)
   zn_cursor_update_position(self, new_screen, screen_x, screen_y);
 }
 
+void
+zn_cursor_set_surface(struct zn_cursor* self, struct wlr_surface* surface)
+{
+  self->texture = wlr_surface_get_texture(surface);
+}
+
 struct zn_cursor*
 zn_cursor_create(void)
 {
--- a/zen/cursor.c
+++ b/zen/cursor.c
@@ -118,6 +118,7 @@ zn_cursor_set_surface(struct zn_cursor* self, struct wlr_surface* surface)
     wl_signal_add(&surface->events.destroy, &self->destroy_surface_listener);
   }
 
+  self->visible = surface != NULL;
   self->surface = surface;
 }
 
--- a/zen/cursor.c
+++ b/zen/cursor.c
@@ -98,7 +98,11 @@ zn_cursor_move_relative(struct zn_cursor* self, int dx, int dy)
 void
 zn_cursor_set_surface(struct zn_cursor* self, struct wlr_surface* surface)
 {
-  self->texture = wlr_surface_get_texture(surface);
+  if (surface == NULL) {
+    self->texture = NULL;
+  } else {
+    self->texture = wlr_surface_get_texture(surface);
+  }
 }
 
 struct zn_cursor*
--- a/zen/cursor.c
+++ b/zen/cursor.c
@@ -73,6 +73,17 @@ zn_cursor_handle_destroy_screen(struct wl_listener* listener, void* data)
       self, found ? screen : NULL, box.width / 2, box.height / 2);
 }
 
+static void
+zn_cursor_handle_destroy_surface(struct wl_listener* listener, void* data)
+{
+  UNUSED(data);
+  struct zn_cursor* self =
+      zn_container_of(listener, self, destroy_surface_listener);
+
+  wl_list_remove(&self->destroy_surface_listener.link);
+  self->surface = NULL;
+}
+
 void
 zn_cursor_move_relative(struct zn_cursor* self, int dx, int dy)
 {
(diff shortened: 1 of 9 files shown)


Aki-7:
I'll look into it now.

First, I want to ask you some current behavior.

- Will you correct cursor hotspot later? If so, it's ok.
- The cursor will not change in xwayland app, is this what you exepcted? I guess we should implement both in the same PR.

watasuke102:
1. fix hotspot issue (#100) is already fixed, but it will be other PR.
2. I forgot to fix

Aki-7:
> I forgot to fix

ok, let me know when fixed. I'll look into this then.

zen/seat.c
```diff
@@ -5,6 +5,16 @@
...
 
+static void
+zn_seat_handle_request_set_cursor(struct wl_listener* listener, void* data)
+{
+  struct zn_seat* self =
+      zn_container_of(listener, self, request_set_cursor_listener);
+  struct wlr_seat_pointer_request_set_cursor_event* event = data;
+
+  zn_cursor_set_surface(self->cursor, event->surface);
+}
```
Aki-7, line 16: Please check that the `self->wlr_seat->pointer_state.focused_client` is the same as `event->seat_client->client`.

See also https://github.com/swaywm/sway/blob/5543acff06981639086bc9a0fc9b608796a23e84/sway/input/cursor.c#L899-L924

zen/output.c
```diff
@@ -71,6 +74,10 @@ zn_output_damage_frame_handler(struct wl_listener *listener, void *data)
 
   wl_list_for_each(view, &self->screen->views, link)
       wlr_surface_send_frame_done(view->impl->get_wlr_surface(view), &now);
+
+  if (cursor->screen == self->screen && cursor->surface != NULL) {
+    wlr_surface_send_frame_done(cursor->surface, &now);
+  }
```
Aki-7, line 80: Which surface is visible and needed to receive frame event are not simple. Those surfaces should be determined by zn_screen not zn_output. So Having a `zn_screen_for_each_visible_surface(iterfation_handler)` would be better.
But you don't have to do this in this PR.

Aki-7 (changes requested):
I found a bug, but you can leave it as an new issue.

Place cursor on the X client, and kill the X client
  - expected: cursor image will get back to the default one.
  - actual: unexpected cursor image are shown until moving the cursor.

Aki-7:
Regarding [this](https://github.com/zigen-project/zen/pull/107#pullrequestreview-1076706795), please leave issue, and link to this PR.

zen/output.c
watasuke102: I want to solve in this PR, but I can't complehend how to implement that. What is `iteration_handler`? What should it be checked?

Aki-7: Pseudo code
```c
typedef void (*zn_screen_surface_iterator_func_t) (struct zn_screen* screen, struct wlr_surface *surface, void *data);

void zn_screen_for_each_visible_surface(struct cn_screen *self, zn_screen_surface_iterator_func_t iterator, void *data) {
  for(view, self->view_list) {
    iterator(self, view->wlr_surface, data);
  }

  if (cursor->screen == self) iterator(self, view->wlr_surface, data);
} 
```

zen/render-2d.c
```diff
@@ -29,15 +29,25 @@ static void
...
+    return;
+  }
+
+  if (cursor->surface != NULL) {
+    texture = wlr_surface_get_texture(cursor->surface);
+  } else {
+    texture = cursor->texture;
+  }
+
+  if (cursor->texture) {
```
Aki-7, line 49: nits
```suggestion
  if (texture) {
```

zen/cursor.c
```diff
@@ -73,6 +73,17 @@ zn_cursor_handle_destroy_screen(struct wl_listener* listener, void* data)
...
 
+static void
+zn_cursor_handle_destroy_surface(struct wl_listener* listener, void* data)
+{
+  UNUSED(data);
+  struct zn_cursor* self =
+      zn_container_of(listener, self, destroy_surface_listener);
+
+  wl_list_remove(&self->destroy_surface_listener.link);
+  self->surface = NULL;
```
Aki-7, line 84:
```suggestion
  zn_cursor_set_surface(self, NULL);
```

zen/cursor.c
```diff
@@ -73,6 +73,17 @@ zn_cursor_handle_destroy_screen(struct wl_listener* listener, void* data)
...
 }
 
+static void
+zn_cursor_handle_destroy_surface(struct wl_listener* listener, void* data)
+{
+  UNUSED(data);
+  struct zn_cursor* self =
+      zn_container_of(listener, self, destroy_surface_listener);
+
+  wl_list_remove(&self->destroy_surface_listener.link);
```
Aki-7, line 83:
```suggestion
```

Aki-7 (commented):
https://github.com/zigen-project/zen/pull/107#discussion_r950276174

Do you forget this?

watasuke102:
> Do you forget this?

I fixed, but forgot to push…

Aki-7 (approved):
Good job!!!

PR #109, opened 2022-08-17
board

Aki-7:
## Context

See https://github.com/zigen-project/zen/issues/111

## Summary

- [x] create the concept of board 
  - A screen has many boards
  - A screen must have at least one board
  - A board must be owned by one screen if there are some screen.
  - If there is no screen, all boards must be owned by no screen. (**dangling board**)
  - A boards has many views
  - A mapped view must be owned by one board.
  - A screen must have current board, which is shown to a display.
- [x] handle screen creation/destruction
  - When a screen is created
    - if there is no dangling board (another screen exists), create a new board and assign it to the screen.
    - if there is dangling board, assign one of the dangling board to the screen.
  - When a screen is destroyed
    - if there are screens remaining, assign destroyed screen's boards to the remaining screen.
    - if there is no screen remaining, the boards will be dangling boards. 
- [x] move between boards
  - `<mod> + <shift> + →`: move to the next board on the screen which cursor is on.  
  - `<mod> + <shift> + ←`: move to the previous board on the screen which cursor is on.  
- [x] enable to create new board
- [x] update README

## How to check behavior

1. build
2. start `zen-desktop`

### Single monitor

3. start a client
an application will be shown
4. create a new board `<mod> + <shift> + n`
new board will be created and empty board will be shown
5. start another client
an application will be shown
6. move between boards `<mod> + <shift> + ←/→`
you can see an application previously created

### Multi monitor

3. Connect multiple monitor (or you have already done this)
4. move cursor to one of your screen
5. start a client
the client will appear on the screen which the cursor is on.
6. move the cursor to another screen, and start a client
the client will appear on the screen which the cursor is on.
7. remove all but one screen
removed screen's boards will be moved to the remaining one board.
9. move between boards `<mod> + <shift> + ←/→`
you can see all boards.

### Find bugs!

remove all screen, start client while no output connected, create new board while no output connected, etc...

### DIFF
--- a/zen/scene/board.c
+++ b/zen/scene/board.c
@@ -1,0 +1,99 @@
+#include "zen/scene/board.h"
+
+#include "zen-common.h"
+
+static void
+zn_board_screen_destroy_handler(struct wl_listener *listener, void *data)
+{
+  struct zn_board *self =
+      zn_container_of(listener, self, screen_destroy_listener);
+  UNUSED(data);
+
+  zn_board_assign_to_screen(self, NULL);
+}
+
+static void
+zn_board_resize(struct zn_board *self, double width, double height)
+{
+  self->width = width;
+  self->height = height;
+  // TODO: move views inside the resized board
+}
+
+bool
+zn_board_is_dangling(struct zn_board *self)
+{
+  return self->screen == NULL;
+}
+
+void
+zn_board_assign_to_screen(struct zn_board *self, struct zn_screen *screen)
+{
+  struct zn_board_screen_assigned_event event;
+  if (self->screen == screen) {
+    return;
+  }
+
+  if (!zn_board_is_dangling(self)) {
+    wl_list_remove(&self->screen_link);
+    wl_list_init(&self->screen_link);
+
+    wl_list_remove(&self->screen_destroy_listener.link);
+    wl_list_init(&self->screen_destroy_listener.link);
+  }
+
+  if (screen) {
+    struct wlr_box box;
+
+    wl_list_insert(screen->board_list.prev, &self->screen_link);
+    wl_signal_add(&screen->events.destroy, &self->screen_destroy_listener);
+
+    zn_screen_get_box(screen, &box);
+    zn_board_resize(self, box.width, box.height);
+  }
+
+  event.prev_screen = self->screen;
+  event.current_screen = screen;
+
+  self->screen = screen;
+
+  wl_signal_emit(&self->events.screen_assigned, &event);
+}
+
+struct zn_board *
+zn_board_create(struct zn_scene *scene)
+{
+  struct zn_board *self;
+
+  self = zalloc(sizeof *self);
+  if (self == NULL) {
+    zn_error("Failed to allocate memory");
+    goto err;
+  }
+
+  self->width = 1920;
+  self->height = 1080;
+
+  self->scene = scene;
+  wl_list_init(&self->view_list);
+  wl_list_insert(scene->board_list.prev, &self->link);
+  wl_list_init(&self->screen_link);
+  self->screen = NULL;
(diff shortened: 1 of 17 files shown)


watasuke102:
I just tried (not yet to read the codes) this each single and multiple screen, and it looks work well.

- remove all screen
  - no problem
- start client while no output connected
  - crashed zen when launching either wayland or X client, because `zn_screen_get_current_board()` receives NULL
- create new board while no output connected
  - cannot create

And I'm using outputs which has different resolution (1280x720, 1280x1024). Place client to 1280x1024 output, and disconnect this output, then the board  (has the client) is moved to 1280x720 output, but the client's position is shifted down. The difference of position is not so large in this case, but if remove big output, then the client might be placed out of screen.

shierote:
In my environment, it seems to work file.
It passed all behavior checks!
Is deleting board is future work?

Aki-7:
Yes, and it's an issue open to everyone.

zen/scene/view.c
```diff
@@ -21,22 +22,35 @@ zn_view_get_fbox(struct zn_view *self, struct wlr_fbox *fbox)
...
 zn_view_is_mapped(struct zn_view *self)
 {
-  // some zn_screen has this view in zn_screen::views
+  // check if some zn_board has this view in zn_board::view_list
   return !wl_list_empty(&self->link);
 }
 
 void
-zn_view_map_to_screen(struct zn_view *self, struct zn_screen *screen)
+zn_view_map_to_scene(struct zn_view *self, struct zn_scene *scene)
```
watasuke102, line 30: I think map_to_board is better.

zen/scene/board.c
```diff
@@ -0,0 +1,99 @@
...
+err:
+  return NULL;
+}
+
+void
+zn_board_destroy(struct zn_board *self)
+{
+  zn_board_assign_to_screen(self, NULL);
+  wl_list_remove(&self->view_list);
+  wl_list_remove(&self->link);
```
shierote, line 97: call other destroy process such as `wl_list_remove(screen_link)`?

zen/scene/board.c
```diff
@@ -0,0 +1,99 @@
...
+
+  self->width = 1920;
+  self->height = 1080;
+
+  self->scene = scene;
+  wl_list_init(&self->view_list);
+  wl_list_insert(scene->board_list.prev, &self->link);
+  wl_list_init(&self->screen_link);
+  self->screen = NULL;
+  self->screen_destroy_listener.notify = zn_board_screen_destroy_handler;
```
shierote, line 82: A few line breaks make it easier to read!

shierote (approved):
LGTM
I wrote a few nits comments.

zen/scene/view.c
Aki-7: I want the `scene` to be responsible for the UX parts that are likely to change, such as how to switch boards, which boards to add an new view for, etc.

zen/scene/board.c
Aki-7: `zn_board_assign_to_screen` does.

watasuke102 (approved):
**_LGTM_**

zen/scene/screen.c
```diff
@@ -45,6 +48,134 @@ zn_screen_get_view_at(
...
 }
 
+void
+zn_screen_switch_to_next_board(struct zn_screen *self)
+{
+  struct zn_board *board, *next_board = NULL,
+                          *current_board = zn_screen_get_current_board(self);
+  bool found = false;
+
+  wl_list_for_each(board, &self->board_list, screen_link)
```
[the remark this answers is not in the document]
Aki-7: I couldn't understand what you mean by `check`, but probably not possible, given that current_board is the first or last element, or that current_board is not included in the board_list.

PR #114, opened 2022-08-20
fix X client could not launch

watasuke102:
## Context

After commit 3489290 (by #107), zen is crashed with SIGSEGV when the X client is launched.

## Summary

NULL check at `zn_seat_handle_request_set_cursor()`

## How to check behavior

1. run zen-desktop with -s option (launch X client)
2. work without terminating


watasuke102:
fixed #115

PR #115, opened 2022-08-20
handle when focus_surface is NULL

Aki-7:
## Context

Bug fix.

When starting X client, the compositor crushes.

## Summary

Fixed the bug.

## How to check behavior

Same as https://github.com/zigen-project/zen/pull/107.


Aki-7:
I'll merge this without review as it's a small bug-fix change.

PR #119, opened 2022-08-22
fix #96

### LINKED ISSUE #96: remove screen in screen layout
When screen is removed from screen_layout, remaining screen's position should be updated, but not.

watasuke102:
## Context

See #96

## Summary

Separate screen arrangement process to new function.

## How to check behavior

None


Aki-7 (approved):
LGTM!

PR #108, opened 2022-08-17
show wayland client's popup/subsurface

shierote:
## Context

https://github.com/zigen-project/zen/issues/116

## Summary

In this PR, I focused on showing/hiding popup and subsurface.

- [x] show popup surface
- [x] hide popup surface
- [x] show subsurface
- [x] hide subsurface

Another features related to popup/subsurface yet not implemented are described in issue #116 #110  and it will be implemented soon.

## How to check behavior

1. run zen-desktop with a wayland client such as weston-terminal.
2. right-click on the wayland client.
3. popup appears near cussor.
4. click another area, and the popup disappears.

PR #121, opened 2022-08-22
damage

Aki-7:
## Context

Damage is the area of an image buffer that has changed.
By handling damage precisely, we can reduce repainting cost greatly.

## Summary

- [x] use effective / transformed coordinates neatly
- [x] handle cursor damage
- [x] handle view damage

## Leaving Issues

- #122 
implement precise view damage handling. We will do this after merging #108.
- #123
enable to change output scale / transform (at least for behavior checking during development)

## How to check behavior

No visible change is made.

Check behavior by modifying the code.

### Check output scale / transform

1. In `output.c#wlr_output_create`, between `wlr_output_enable` and `if (wlr_output_commit(..) == false) {`, insert 
```c
  wlr_output_set_transform(self->wlr_output, WL_OUTPUT_TRANSFORM_90);
  wlr_output_set_scale(self->wlr_output, 0.5);
```
2. build
3. run `zen-desktop`

You can use the output by rotate it 90 degrees clockwise.

### Check damage

1. Make `view.c@zn_view_damage_whole` empty.
```c
void
zn_view_damage_whole(struct zn_view *self)
{
  UNUSED(self);
}
```
This will stop notifying view's damage.

2. build
3. run `zen-desktop`
4. run some client.
5. move the cursor around.

You can see the client only the part the cursor has passed, which means that only the region cursor has passed is repainted, and rest region of the screen is not repainted.

### DIFF
--- a/include/zen/cursor.h
+++ b/include/zen/cursor.h
@@ -11,19 +11,24 @@
 
 struct zn_cursor {
   double x, y;
+  uint32_t width, height;
   int hotspot_x, hotspot_y;
   bool visible;
+
   struct zn_screen* screen;  // nullable
   struct wlr_texture* texture;
   struct wlr_surface* surface;
   struct wlr_xcursor_manager* xcursor_manager;
 
   struct wl_listener new_screen_listener;
-  struct wl_listener destroy_screen_listener;
-  struct wl_listener destroy_surface_listener;
+  struct wl_listener screen_destroy_listener;
+  struct wl_listener surface_commit_listener;
+  struct wl_listener surface_destroy_listener;
 };
 
 void zn_cursor_move_relative(struct zn_cursor* self, double dx, double dy);
+
+void zn_cursor_get_fbox(struct zn_cursor* self, struct wlr_fbox* fbox);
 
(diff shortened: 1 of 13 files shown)


watasuke102:
- output scale / transform
  - I tried both, and looks work well.
- damage
  - I tried with `weston-clickdot` and `weston-terminal`, but they alternately rendered the area the cursor has passed and the area all of the view. It switched when the cursor is moved even outside of the view.

Aki-7:
> I tried with weston-clickdot and weston-terminal with -s option, but they alternately rendered the area the cursor has passed and the area all of the view. It switched when the cursor is moved even outside of the view.

It happens when making view.c@zn_view_damage_whole empty, right?

watasuke102:
> It happens when making view.c@zn_view_damage_whole empty

Yes, I added `return` to the top of `zn_view_damage_whole()`.

Aki-7:
Then, I think that's no problem. A function damaging whole output (`wlr_output_damage_whole` or `wlr_output_damage_add_whole`) was called at somewhere in your environment. That could happen.

Aki-7:
conflict resolved

watasuke102 (approved):
Nice change!

PR #124, opened 2022-08-24
Focus view

watasuke102:
## Context

Focus / unfocus to view. So far, change is only that it appears brighter than the other views.

## Summary

Using `wlr_xdg_toplevel_set_activated` and `wlr_xwayland_surface_activate`

## How to check behavior

1. start zen more than 2 clients
2. click each client; the view is activated
3. click out of the view; the view is inactivated

### DIFF
--- a/include/zen/scene/board.h
+++ b/include/zen/scene/board.h
@@ -27,7 +27,6 @@ struct zn_board {
   struct zn_screen *screen;    // nullable, screen which this board belongs.
 
   struct wl_listener screen_destroy_listener;
-  struct wl_listener focus_listener;
 
   struct {
     struct wl_signal
--- a/include/zen/scene/board.h
+++ b/include/zen/scene/board.h
@@ -27,6 +27,7 @@ struct zn_board {
   struct zn_screen *screen;    // nullable, screen which this board belongs.
 
   struct wl_listener screen_destroy_listener;
+  struct wl_listener focus_listener;
 
   struct {
     struct wl_signal
--- a/include/zen/scene/scene.h
+++ b/include/zen/scene/scene.h
@@ -8,6 +8,8 @@
 struct zn_scene {
   struct zn_screen_layout* screen_layout;
 
+  struct zn_view* active_view;
+
   struct wl_list board_list;  // zn_board::link, non empty
 };
 
--- a/include/zen/scene/scene.h
+++ b/include/zen/scene/scene.h
@@ -12,7 +12,7 @@ struct zn_scene {
 
   struct wl_list board_list;  // zn_board::link, non empty
 
-  struct wl_listener destroy_focused_view_listener;
+  struct wl_listener unmap_focused_view_listener;
 };
 
 void zn_scene_set_focused_view(struct zn_scene* self, struct zn_view* view);
--- a/include/zen/scene/scene.h
+++ b/include/zen/scene/scene.h
@@ -11,7 +11,11 @@ struct zn_scene {
   struct zn_view* active_view;
 
   struct wl_list board_list;  // zn_board::link, non empty
+
+  struct wl_listener destroy_active_view_listener;
 };
+
+void zn_scene_set_active_view(struct zn_scene* self, struct zn_view* view);
 
 struct zn_board* zn_scene_get_focus_board(struct zn_scene* self);
 
--- a/include/zen/scene/scene.h
+++ b/include/zen/scene/scene.h
@@ -8,14 +8,14 @@
 struct zn_scene {
   struct zn_screen_layout* screen_layout;
 
-  struct zn_view* active_view;
+  struct zn_view* focused_view;
 
   struct wl_list board_list;  // zn_board::link, non empty
 
-  struct wl_listener destroy_active_view_listener;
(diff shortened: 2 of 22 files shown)


include/zen/scene/board.h
```diff
@@ -27,6 +27,7 @@ struct zn_board {
   struct zn_screen *screen;    // nullable, screen which this board belongs.
 
   struct wl_listener screen_destroy_listener;
+  struct wl_listener focus_listener;
```
Aki-7, line 30: unused

include/zen/scene/scene.h
```diff
@@ -8,9 +8,15 @@
 struct zn_scene {
   struct zn_screen_layout* screen_layout;
 
+  struct zn_view* active_view;
+
   struct wl_list board_list;  // zn_board::link, non empty
+
+  struct wl_listener destroy_active_view_listener;
```
Aki-7, line 15: Should listen `unmap`. Unmap always happens before beeing destroyed.

zen/scene/view.c
```diff
@@ -26,6 +26,36 @@ zn_view_is_mapped(struct zn_view *self)
...
+  struct zn_server *server = zn_server_get_singleton();
+  struct zn_scene *scene = server->scene;
+
+  if (scene->active_view != NULL) {
+    zn_view_unfocus(scene->active_view);
+  }
+
+  if (self != NULL) {
+    self->impl->focus(self);
+  }
```
Aki-7, line 41: We should not assume `self` is NULL
```suggestion
  self->impl->focus(self);
```

zen/scene/view.c
```diff
@@ -26,6 +26,36 @@ zn_view_is_mapped(struct zn_view *self)
...
+
+void
+zn_view_unfocus(struct zn_view *self)
+{
+  struct zn_server *server = zn_server_get_singleton();
+  struct zn_scene *scene = server->scene;
+
+  if (self != NULL) {
+    self->impl->unfocus(self);
+  }
```
Aki-7, line 54:
```suggestion
  self->impl->unfocus(self);
```

zen/input/pointer.c
```diff
@@ -47,10 +47,23 @@ zn_pointer_handle_button(struct wl_listener* listener, void* data)
...
+
+  if (event->state == WLR_BUTTON_PRESSED) {
+    view =
+        zn_screen_get_view_at(cursor->screen, cursor->x, cursor->y, NULL, NULL);
+    if (view == NULL) {
+      zn_view_unfocus(active_view);
+    } else {
+      zn_view_focus(view);
+    }
+  }
```
Aki-7, line 66: Sway do this before `wlr_seat_pointer_notify_button`. Is there any reason to change the order?

include/zen/scene/scene.h
```diff
@@ -8,9 +8,15 @@
 struct zn_scene {
   struct zn_screen_layout* screen_layout;
 
+  struct zn_view* active_view;
```
Aki-7, line 11:
```suggestion
  struct zn_view* focus_view;
```

Do we have to use different term, `active` and `focus`?

zen/input/pointer.c
Aki-7: I think it's more safe to call `zn_scene_set_active_view` here and call `zn_view_(focus|unfocus)` inside the `zn_scene_set_active_view`.

zen/xwayland-view.c
```diff
@@ -44,6 +44,20 @@ zn_xwayland_view_wlr_xwayland_surface_destroy_handler(
   zn_xwayland_view_destroy(self);
 }
 
+static void
+zn_xwayland_view_impl_focus(struct zn_view* view)
+{
+  struct zn_xwayland_view* self = zn_container_of(view, self, base);
+  wlr_xwayland_surface_activate(self->wlr_xwayland_surface, true);
```
Aki-7, line 51:
```suggestion
  wlr_xwayland_surface_activate(self->wlr_xwayland_surface, true);
  wlr_xwayland_surface_restack(surface, NULL, XCB_STACK_MODE_ABOVE);
```

We don't need `wlr_xwayland_surface_restack`?

include/zen/scene/view.h
```diff
@@ -11,6 +11,8 @@ struct zn_view;
 
 struct zn_view_impl {
   struct wlr_surface *(*get_wlr_surface)(struct zn_view *view);
+  void (*focus)(struct zn_view *view);
+  void (*unfocus)(struct zn_view *view);
```
Aki-7, line 15:
```suggestion
  void (*set_activated)(struct zn_view *view, bool activated);
```

I think it is more in line with the actual situation to think of `activate` as one of the processes of `focus`.
`focus` has other processes such as reordering the views.

A single method with a boolean value is more simple, I think.

Aki-7 (changes requested):

zen/xwayland-view.c
watasuke102: I'm going to implement about restack as another PR.

Aki-7:
Please resolve the conflicts, then I'll look into this.

watasuke102:
close; this feature is implemented by #134

PR #128, opened 2022-08-26
Preliminary refactoring to create Immersive Renderer

Aki-7:
## Context

Refactoring for immersive renderer

## Summary

- [x] rename `zn_render_2d` to `zn_screen_renderer`
  - We decided to call `screen` and `immersive` display system, not `2d` and `3d`. So we should rename `zn_render_2d` before creating `zn_screen_renderer`.
- [x] stop separating immersive backend from core components. 
  - In other words, move `immersive-backend/` to `zen/backend/`
  - In general, peripheral components should be dependent on the central component. In our case, `zen/scene/` would be the most central, outside of it would be the other objects under `zen/`, and this immersive backend would be further outside of it. We will now create objects for immersive display system (e.g., `zn_camera`). Some of them should be created by immersive backend, and thus immersive backend should know core components. That's the reason of this change.

## How to check behavior

No behavioral update


shierote (approved):
good!

watasuke102 (approved):

PR #129, opened 2022-08-27
fix #126

watasuke102:
## Context

In X client, the cursor event looks to be sent to wrong view.

See #126 

the cursor event looks to be sent to wrong view

## Summary

Call `wlr_xwayland_surface_configure()` when the view is moved.

the cursor event looks to be sent to correct view

## How to check behavior

1. start zen with more than two of X client
2. move views somehow
3. move the cursor

### move views (recommend; OPTIONAL)

remove `static` from `zn_view_move`, create function like following:

```c
static void
zn_scene_rearrange_view(uint32_t time_msec, uint32_t key, void* data)
{
  UNUSED(time_msec);
  UNUSED(key);
  UNUSED(data);
  extern void zn_view_move(struct zn_view* self, double x, double y);
  struct zn_board* board = zn_scene_get_focus_board(zn_server_get_singleton()->scene);
  struct zn_view* view;
  struct wlr_fbox box;
  double x = 0;

  wl_list_for_each(view, &board->view_list, link)
  {
    zn_view_get_fbox(view, &box);
    zn_view_damage_whole(view);
    zn_view_move(view, x, 0);
    zn_view_damage_whole(view);
    x += box.width - 100;
  }
}
```

and add key binding:

```c
  zn_input_manager_add_key_binding(
      server->input_manager, KEY_R, WLR_MODIFIER_LOGO, zn_scene_rearrange_view, self);
```

PR #130, opened 2022-08-28
Use correct view box

watasuke102:
## Context

The event isn't sent near the border of the view.

image

## Summary

Add `zn_view_get_window_fbox()`, and use this at `zn_screen_get_view_at()`.

## How to check behavior

(Same with #129)

1. start zen with more than two of X client
2. move views somehow
3. move the cursor

### DIFF
--- a/include/zen/cursor.h
+++ b/include/zen/cursor.h
@@ -13,7 +13,6 @@ struct zn_cursor {
   double x, y;
   uint32_t width, height;
   int hotspot_x, hotspot_y;
-  int default_hotspot_x, default_hotspot_y;
   bool visible;
 
   struct zn_screen* screen;  // nullable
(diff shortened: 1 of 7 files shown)


Aki-7 (approved):
Nice!

Aki-7:
We must also consider the input region in the future.
ref
- https://wayland.app/protocols/wayland#wl_surface:request:set_input_region

To implement this, we must handle subsurface and popup properly.

Aki-7:
See also
- https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/0.15/types/xdg_shell/wlr_xdg_surface.c#L531-540
- https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/0.15/types/wlr_surface.c#L1183-1227

PR #132, opened 2022-08-29
Use default hotspot

watasuke102:
## Context

When the cursor's surface is NULL, default texture is used on rendering, but it use wrong hotspot.

## Summary

Set default hotspot when the cursor's surface is NULL.

## How to check behavior

None

### DIFF
--- a/include/zen/cursor.h
+++ b/include/zen/cursor.h
@@ -13,6 +13,7 @@ struct zn_cursor {
   double x, y;
   uint32_t width, height;
   int hotspot_x, hotspot_y;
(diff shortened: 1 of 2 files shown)


Aki-7 (approved):
Considering that we would have to display an xcursor other than `left_ptr` when resizing, etc., I thought it would be better to set `left_ptr` xcursor every time the cursor goes out of windows, rather than storing a default xcursor texture or hotspot. Then we don't need to store default_hotspot as well.

You can merge this, and leave it as a future work.

PR #134, opened 2022-08-29
Focus to view with grab

watasuke102:
## Context

Focus / unfocus to view

## Summary

Almost same with #124, but using grab.

## How to check behavior

1. start zen more than 2 clients
2. click each client; the view is activated
3. click out of the view; the view is inactivated

(Same with #124)


Aki-7 (changes requested):
Let me summarize a few concepts.

`zn_pointer` represents a pointing device, and `zn_cursor` represents a cursor shown on a display. When you have multiple pointing devices, multiple `zn_pointer` control `zn_cursor`.
In other words, `zn_poitner` represents some hardware, and `zn_cursor` represents the 
conceptual cursor you can see on displays.
And `grab` is an object that represents the behavior of the cursor, not the behavior of each physical pointing device.

So, you should have `zn_cursor_grab` (not `zn_pointer_grab`). Practically, otherwise you need to change all `zn_pointer`'s grab when switching grab.

(That's also the reason why it's wrong to call `zn_screen_get_at_view` in `zn_pointer_handle_motion`)

watasuke102:
I forgot that multiple pointers may be created...

zen/input/pointer.c
```diff
@@ -75,7 +54,7 @@ zn_pointer_handle_frame(struct wl_listener* listener, void* data)
   UNUSED(data);
   struct zn_server* server = zn_server_get_singleton();
   struct wlr_seat* seat = server->input_manager->seat->wlr_seat;
-  wlr_seat_pointer_notify_frame(seat);
+  wlr_seat_pointer_send_frame(seat);
```
Aki-7, line 57: Are you going to add `frame` and `axis` to grab interface?
I think you can do so within this PR (not so big changes will be added).

zen/input/pointer.c
```diff
@@ -13,48 +13,27 @@ static void
...
   UNUSED(listener);
-  struct wlr_event_pointer_button* event = data;
   struct zn_server* server = zn_server_get_singleton();
-  struct wlr_seat* seat = server->input_manager->seat->wlr_seat;
+  struct zn_cursor* cursor = server->input_manager->seat->cursor;
+  struct wlr_event_pointer_button* event = data;
 
-  wlr_seat_pointer_notify_button(
-      seat, event->time_msec, event->button, event->state);
+  cursor->grab->interface->button(cursor->grab, event);
```
Aki-7, line 36: How about creating this kind of methods?

```c
void
zn_cursor_notify_button(struct zn_cursor *self, struct wlr_event_pointer_button *event) {
  self->grab->interface->button(self->grab, event);
}
```

I just think it's more beautiful that way, not mandatory.

include/zen/cursor.h
```diff
@@ -2,9 +2,11 @@
 #define ZEN_CURSOR_H
 
 #include <wayland-server.h>
+#include <wlr/interfaces/wlr_pointer.h>
```
Aki-7, line 5: This seems a needless change.

include/zen/scene/view.h
```diff
@@ -15,6 +15,7 @@ struct zn_view_impl {
   void (*configure)(struct zn_view *view, double x, double y);
   void (*for_each_popup_surface)(struct zn_view *view,
       wlr_surface_iterator_func_t iterator, void *user_data);
+  void (*set_activated)(struct zn_view *view, bool activate);
```
Aki-7, line 18: Please add comments which tell us which one is nullable and which one is not.

zen/cursor.c
```diff
@@ -8,9 +8,66 @@
...
+  struct wlr_seat* seat = server->input_manager->seat->wlr_seat;
+  struct zn_cursor* cursor = server->input_manager->seat->cursor;
+  struct zn_view* view;
+
+  wlr_seat_pointer_send_button(
+      seat, event->time_msec, event->button, event->state);
+
+  if (event->state == WLR_BUTTON_PRESSED) {
+    view =
+        zn_screen_get_view_at(cursor->screen, cursor->x, cursor->y, NULL, NULL);
```
Aki-7, line 61: `cursor->screen` can be NULL.

Aki-7 (changes requested):

zen/input/pointer.c
watasuke102: Why did you think it is beautiful?

zen/input/pointer.c
watasuke102: I won't add `axis` interface, only add `frame` for later implements. If you think it isn't required yet, I'll remove `frame` interface.

Aki-7: I think we need both. Some grab may not want to send axis event.

zen/input/pointer.c
Aki-7: `self->grab->interface->button(self->grab, event);`

Every time we use grab interface, we need to type `self->grab` twice. That's a little bit bothering for me.
And we are doing something interesting by using this grab, changing the behavior of the same function call from time to time (transparency), but I think that `self->grab->interface->button` shows a transparent mechanism too much and loses the fun of transparency.

watasuke102: For now, `zn_cursor_notify_button` will just be an alias to `grab->interface->button`, and it seems never change. When reading the code, such redundant statements seem to create unnecessary costs (I felt this when reading wlroots).

This cost is more than the cost to writing `self->grab` twice, I think.

Aki-7: Good insight! Let's keep it that way.

zen/input/pointer.c
```diff
@@ -13,48 +13,27 @@ static void
...
-  }
-
-  if (surface) {
-    wlr_seat_pointer_notify_enter(seat, surface, view_x, view_y);
-    wlr_seat_pointer_notify_motion(seat, event->time_msec, view_x, view_y);
-  } else {
-    zn_cursor_reset_surface(cursor);
-    wlr_seat_pointer_notify_clear_focus(seat);
-  }
+  cursor->grab->interface->motion(cursor->grab, event);
```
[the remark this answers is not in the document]
Aki-7: > Is there any reason to callzn_cursor_move_relative outside of cursor->grab->interface->motion?

How about this?

include/zen/scene/view.h
```diff
@@ -15,6 +15,7 @@ struct zn_view_impl {
   void (*configure)(struct zn_view *view, double x, double y);
   void (*for_each_popup_surface)(struct zn_view *view,
       wlr_surface_iterator_func_t iterator, void *user_data);
+  void (*set_activated)(struct zn_view *view, bool activate);  // nullable
```
Aki-7, line 18: `nullable`? Then you need to do null check in `scene.c@zn_scene_set_focused_view`.

zen/cursor.c
```diff
@@ -135,6 +211,18 @@ zn_cursor_handle_surface_destroy(struct wl_listener* listener, void* data)
...
+{
+  self->grab = grab;
+}
+
+void
+zn_cursor_end_grab(struct zn_cursor* self)
+{
+  self->grab = &self->grab_default;
+}
+
```
Aki-7, line 225: I cannot review these functions without using them.

Please add them in another PR.

zen/input/pointer.c
Aki-7: Why did you add only `axis`? How about `frame`?

Aki-7 (changes requested):

Aki-7:
https://github.com/zigen-project/zen/pull/134#discussion_r958324339

zen/input/pointer.c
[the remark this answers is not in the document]
watasuke102: I think there is not the situation that the cursor doesn't move when the pointer received the motion event.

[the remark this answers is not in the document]
Aki-7: > I think there is not the situation that the cursor doesn't move when the pointer received the motion event.
That's true, but this is not a positive reason. One positive reason to choose this approach is that sway takes similar approach, so this may work as well. But be careful about the following limitation.

- How the cursor moves.
  - You cannot change cursor motion behavior while move, resize, dnd, etc.
- We cannot use previous position of the cursor in grab.
  - Weston use previous surface local position in a default grab (I don't know what it's for).
- It's more difficult to decide what process should be written in `zn_pointer_handle_motion` and what is in `grab->interface->motion`

Let's this discussion be resolved. You don't need to fix this.

Aki-7 (approved):
Nice work!

Issue #136, opened 2022-08-30
Damage whole when switching board


Aki-7:

Aki-7:
same as #131

PR #137, opened 2022-08-30
damage whole screen when switching board

Aki-7:
## Context

#136 

## Summary

- [x] add damage to whole output when switching board.

## How to check behavior

1. Check that views stay displayed when switching board.
2. build
3. start `zen-desktop` 
4. Check that views do not stay displayed when switching board.


watasuke102 (approved):
LGTM

PR #138, opened 2022-08-30
Refactoring VrSystem

Aki-7:
## Context

With previous implementation, we call the OpenVR API across multiple threads.
Whether it works or not depends on the implementation of OpenVR, which is unknown for us.

During implementing immersive renderer, I found it doesn't work, so I decided to use OpenVR API in a single thread.

Instead, the OpenGL context must be switched between threads. (This is because textures created in the main thread need to be submitted in the OpenVR thread).

## Summary

- [x] Stop using OpenVR API across multiple thread.

## How to check behavior

No behavioral change.

### DIFF
--- a/zen/backend/openvr/meson.build
+++ b/zen/backend/openvr/meson.build
@@ -1,5 +1,6 @@
 _zen_openvr_immersive_backend_srcs =[
   'immersive-backend.cc',
+  'openvr.cc',
   'vr-system.cc',
 ]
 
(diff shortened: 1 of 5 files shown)


watasuke102 (approved):

Aki-7:
I'll merge this after implementing immersive renderer and confirming this works well.

Aki-7:
This doesn't work well.

Issue #112, opened 2022-08-18
appear wrong cursor when X client terminate


watasuke102:
I found a bug, but you can leave it as an new issue.

Place cursor on the X client, and kill the X client
  - expected: cursor image will get back to the default one.
  - actual: unexpected cursor image are shown until moving the cursor.

_Originally posted by @Aki-7 in https://github.com/zigen-project/zen/pull/107#pullrequestreview-1076706795_

PR #139, opened 2022-08-31
fix #133

### LINKED ISSUE #133: Use xcursor instead of texture
On #132:

> Considering that we would have to display an xcursor other than having left_ptr when resizing, etc., I thought it would be better to set left_ptr xcursor every time the cursor goes out of windows, rather than storing a default xcursor texture or hotspot. Then we don't need to store default_hotspot as well.

(Use `wlr_xcursor_manager_get_xcursor()`)

watasuke102:
## Context

See #133

## Summary

Added `zn_cursor_set_xcursor()`, and remove cursor::visible from zn_cursor. The visiblity is expressed by whether or not cursor::texture is NULL.

## How to check behavior

None

### DIFF
--- a/zen/cursor.c
+++ b/zen/cursor.c
@@ -41,7 +41,6 @@ default_grab_motion(
     wlr_seat_pointer_enter(seat, surface, view_x, view_y);
     wlr_seat_pointer_send_motion(seat, event->time_msec, view_x, view_y);
   } else {
-    zn_cursor_set_surface(cursor, NULL, 0, 0);
     zn_cursor_set_xcursor(cursor, "left_ptr");
     wlr_seat_pointer_clear_focus(seat);
   }
@@ -214,7 +213,6 @@ zn_cursor_handle_surface_destroy(struct wl_listener* listener, void* data)
   struct zn_cursor* self =
       zn_container_of(listener, self, surface_destroy_listener);
 
-  zn_cursor_set_surface(self, NULL, 0, 0);
   zn_cursor_set_xcursor(self, "left_ptr");
 }
 
@@ -298,9 +296,7 @@ zn_cursor_set_xcursor(struct zn_cursor* self, char* name)
   struct wlr_xcursor* xcursor;
   struct wlr_xcursor_image* image;
 
-  if (self->texture && !self->surface) {
-    wlr_texture_destroy(self->texture);
-  }
+  zn_cursor_set_surface(self, NULL, 0, 0);
 
   xcursor = wlr_xcursor_manager_get_xcursor(self->xcursor_manager, name, 1.f);
   if (xcursor == NULL) {
--- a/zen/cursor.c
+++ b/zen/cursor.c
@@ -98,6 +98,21 @@ static const struct zn_cursor_grab_interface default_grab_interface = {
     .frame = default_grab_frame,
 };
 
+static void
+zn_cursor_update_size(struct zn_cursor* self)
+{
+  if (!self->texture) {
+    self->width = 0;
+    self->height = 0;
+  } else if (self->surface) {
+    self->width = self->surface->current.width;
+    self->height = self->surface->current.height;
+  } else {
+    self->width = self->texture->width;
+    self->height = self->texture->height;
+  }
+}
+
 // screen_x and screen_y must be less than screen->width/height
 static void
 zn_cursor_update_position(struct zn_cursor* self, struct zn_screen* screen,
@@ -189,6 +204,8 @@ zn_cursor_handle_surface_commit(struct wl_listener* listener, void* data)
 
   zn_cursor_damage_whole(self);
 
+  zn_cursor_update_size(self);
+
   zn_cursor_damage_whole(self);
 }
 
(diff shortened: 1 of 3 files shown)


watasuke102:
This PR doesn't have enough consideration

zen/cursor.c
```diff
@@ -290,14 +292,30 @@ zn_cursor_set_surface(struct zn_cursor* self, struct wlr_surface* surface,
 }
 
 void
-zn_cursor_reset_surface(struct zn_cursor* self)
+zn_cursor_set_xcursor(struct zn_cursor* self, char* name)
```
Aki-7, line 295: You suppose that
if (zn_cursor::surface && zn_cursor::texture) : a client set the cursor's surface.
if (zn_cursor::surface && !zn_cursor::texture) : never happen.
if (!zn_cursor::surface && !zn_cursor::texture) : using xcursor
if (!zn_cursor::surface && !zn_cursor::texture) : a client unset cursor's surface (invisible cursor)

Then zn_cursor's public methods should be safe about the consistency above.
Suppose a programmer writes the following code.
```c
zn_cursor_set_surface(cursor, surface, hx, hy);

... later ...

zn_cursor_set_xcursor(cursor, "left_ptr");
```

With current implementation, zn_cursor::surface and ::texture will not be null neither. But doesn't mean using a  surface given by a client.

Your current code works well because you call zn_cursor_set_xcursor and zn_cursor_set_surface at the same time, but it's not safe. Both methods should keep this consistency.

Aki-7 (changes requested):

zen/cursor.c
Aki-7: Thus you don't have to call this two methods every time
```c
  zn_cursor_set_surface(self, NULL, 0, 0);
  zn_cursor_set_xcursor(self, "left_ptr");
```
instead, just
```c
  zn_cursor_set_xcursor(self, "left_ptr");
```

Aki-7:
The code seems ok, but I met an error log.
```
*** BUG ***
In pixman_region32_union_rect: Invalid rectangle passed
Set a breakpoint on '_pixman_log_error' to debug
```

watasuke102:
I can't reproduce that. Can you tell how it occured in detail?

watasuke102:
Can I leave this as another issue? It reproduced in my environment, but it's hardly.

Aki-7:
I found the cause.

zen/cursor.c
```diff
@@ -252,53 +232,78 @@ zn_cursor_move_relative(struct zn_cursor* self, double dx, double dy)
...
+  }
+
   if (self->surface != NULL) {
     wl_list_remove(&self->surface_destroy_listener.link);
     wl_list_init(&self->surface_destroy_listener.link);
     wl_list_remove(&self->surface_commit_listener.link);
     wl_list_init(&self->surface_commit_listener.link);
   }
 
+  zn_cursor_damage_whole(self);
```
Aki-7, line 265: Here, texture has been destroyed but not NULL, and it gets wrong fbox of the cursor.

zen/cursor.c
```diff
@@ -203,12 +187,8 @@ zn_cursor_handle_surface_commit(struct wl_listener* listener, void* data)
     return;
   }
 
-  self->visible = wlr_surface_has_buffer(self->surface);
-
   zn_cursor_damage_whole(self);
 
-  zn_cursor_update_size(self);
-
```
Aki-7, line 211: Here is another problem. Removing this ,  `zn_cursor_damage_whole` will be called in succession.

Originally, the former `zn_cursor_damage_whole` adds a damage of previous surface, and the later one adds a damage of next surface.
We can see from this that we have to keep `zn_cursor::width` and `::height` to add damage of previous surface.

zen/cursor.c
```diff
@@ -252,53 +232,78 @@ zn_cursor_move_relative(struct zn_cursor* self, double dx, double dy)
...
+  image = xcursor->images[0];
 
-  self->visible = true;
+  zn_cursor_damage_whole(self);
 
-  zn_cursor_update_size(self);
+  self->hotspot_x = image->hotspot_x;
+  self->hotspot_y = image->hotspot_y;
+  self->texture = wlr_texture_from_pixels(server->renderer, DRM_FORMAT_ARGB8888,
+      image->width * 4, image->width, image->height, image->buffer);
```
Aki-7, line 306: With this implementation, as `zn_cursor_set_xcursor` will be called every time the cursor moves on the screen where no view exists, `zn_cursor::texture` will be created and destroyed again and again in vain.

As `wlr_xcursor_manager_get_xcursor` returns the same pointer for the same "name", you should check whether the next xcursor is the same as the current xcursor.

Aki-7 (changes requested):
I found a bug, 

Reproduce step,
1. start google-chrome
2. close the google chrome by pressing the closing-tab button.

I submit few reviews, but I review remaining part again after solving this bug.

watasuke102:
probably related #112

Aki-7:
Maybe, but compositor crashing is not acceptable.
If so, please fix #112 first, or within this PR.

watasuke102:
This occured because cursor->texture is different from cursor->surface's texture, so I think before method (cursor->texture contains xcursor's texture, fetch surface's texture every render) is better.

I think rebuilding this PR is may better...

PR #144, opened 2022-09-04
fix #133

### LINKED ISSUE #133: Use xcursor instead of texture

watasuke102:
## Context
See #133 (and this is a rebuilded version of #139)

## Summary
Added `zn_cursor_set_xcursor()`.

## How to check behavior
None


Aki-7 (approved):
LGTM!

PR #145, opened 2022-09-04
Move view

watasuke102:
## Context

Move the view by the cursor.

## Summary

Add `zn_cursor_grab_move`.

## How to check behavior

1. start zen and some wayland client
2. drag the top of the view - the view is moved
3. drag between screens - the view is moved between boards

### DIFF
--- a/include/zen/input/cursor-grab-move.h
+++ b/include/zen/input/cursor-grab-move.h
@@ -8,6 +8,7 @@
 struct zn_cursor_grab_move {
   struct zn_cursor_grab base;
   struct zn_view* view;
+  struct zn_screen* prev_screen;
 
   double diff_x, diff_y;
 };
--- a/include/zen/input/cursor-grab-move.h
+++ b/include/zen/input/cursor-grab-move.h
@@ -12,6 +12,8 @@ struct zn_cursor_grab_move {
 
   double init_x, init_y;
   double diff_x, diff_y;
+
+  struct wl_listener view_unmap_listener;
 };
 
 void zn_cursor_grab_move_start(struct zn_cursor* cursor, struct zn_view* view);
--- a/include/zen/input/cursor-grab-move.h
+++ b/include/zen/input/cursor-grab-move.h
@@ -10,6 +10,7 @@ struct zn_cursor_grab_move {
   struct zn_view* view;
   struct zn_screen* prev_screen;
 
+  struct zn_board* init_board;
   double init_x, init_y;
   double diff_x, diff_y;
 
--- a/include/zen/input/cursor-grab-move.h
+++ b/include/zen/input/cursor-grab-move.h
@@ -1,0 +1,15 @@
+#ifndef ZN_CURSOR_GRAB_MOVE_H
+#define ZN_CURSOR_GRAB_MOVE_H
+
+#include "zen/cursor.h"
+#include "zen/input/cursor-grab.h"
+#include "zen/scene/view.h"
+
+struct zn_cursor_grab_move {
+  struct zn_cursor_grab base;
+  struct zn_view* view;
+};
+
+void zn_cursor_grab_move_start(struct zn_cursor* cursor, struct zn_view* view);
+
+#endif  //  ZN_CURSOR_GRAB_MOVE_H
--- a/include/zen/input/cursor-grab-move.h
+++ b/include/zen/input/cursor-grab-move.h
@@ -14,6 +14,7 @@ struct zn_cursor_grab_move {
   double init_x, init_y;
   double diff_x, diff_y;
 
+  struct wl_listener prev_screen_destroy_listener;
   struct wl_listener view_unmap_listener;
 };
 
--- a/include/zen/input/cursor-grab-move.h
+++ b/include/zen/input/cursor-grab-move.h
@@ -8,13 +8,11 @@
 struct zn_cursor_grab_move {
   struct zn_cursor_grab base;
   struct zn_view* view;
-  struct zn_screen* prev_screen;
 
   struct zn_board* init_board;
   double init_x, init_y;
   double diff_x, diff_y;
 
-  struct wl_listener prev_screen_destroy_listener;
   struct wl_listener view_unmap_listener;
 };
 
--- a/include/zen/input/cursor-grab-move.h
+++ b/include/zen/input/cursor-grab-move.h
@@ -14,6 +14,7 @@ struct zn_cursor_grab_move {
   double diff_x, diff_y;
 
   struct wl_listener view_unmap_listener;
+  struct wl_listener init_board_destroy_listener;
 };
 
 void zn_cursor_grab_move_start(struct zn_cursor* cursor, struct zn_view* view);
--- a/include/zen/input/cursor-grab-move.h
+++ b/include/zen/input/cursor-grab-move.h
@@ -10,6 +10,7 @@ struct zn_cursor_grab_move {
   struct zn_view* view;
   struct zn_screen* prev_screen;
 
+  double init_x, init_y;
   double diff_x, diff_y;
 };
 
--- a/include/zen/input/cursor-grab-move.h
+++ b/include/zen/input/cursor-grab-move.h
@@ -8,6 +8,8 @@
 struct zn_cursor_grab_move {
   struct zn_cursor_grab base;
   struct zn_view* view;
+
+  double diff_x, diff_y;
 };
 
 void zn_cursor_grab_move_start(struct zn_cursor* cursor, struct zn_view* view);
--- a/include/zen/input/cursor-grab-move.h
+++ b/include/zen/input/cursor-grab-move.h
@@ -7,7 +7,7 @@
 
 struct zn_cursor_grab_move {
   struct zn_cursor_grab base;
-  struct zn_view* view;
+  struct zn_view* view;  // nonnull
 
   struct zn_board* init_board;
   double init_x, init_y;
--- a/include/zen/input/cursor-grab.h
+++ b/include/zen/input/cursor-grab.h
@@ -17,6 +17,7 @@ struct zn_cursor_grab_interface {
   void (*axis)(
       struct zn_cursor_grab* grab, struct wlr_event_pointer_axis* event);
   void (*frame)(struct zn_cursor_grab* grab);
+  void (*rebase)(struct zn_cursor_grab* grab);
   void (*cancel)(struct zn_cursor_grab* grab);
 };
 
--- a/include/zen/input/cursor-grab.h
+++ b/include/zen/input/cursor-grab.h
@@ -17,7 +17,6 @@ struct zn_cursor_grab_interface {
   void (*axis)(
       struct zn_cursor_grab* grab, struct wlr_event_pointer_axis* event);
   void (*frame)(struct zn_cursor_grab* grab);
-  // called when zn_cursor::screen is assigned NULL
   void (*cancel)(struct zn_cursor_grab* grab);
 };
 
--- a/include/zen/input/cursor-grab.h
+++ b/include/zen/input/cursor-grab.h
@@ -17,6 +17,7 @@ struct zn_cursor_grab_interface {
   void (*axis)(
       struct zn_cursor_grab* grab, struct wlr_event_pointer_axis* event);
   void (*frame)(struct zn_cursor_grab* grab);
+  // called when zn_cursor::screen is assigned NULL
   void (*cancel)(struct zn_cursor_grab* grab);
 };
 
--- a/include/zen/input/cursor-grab.h
+++ b/include/zen/input/cursor-grab.h
@@ -17,6 +17,7 @@ struct zn_cursor_grab_interface {
   void (*axis)(
       struct zn_cursor_grab* grab, struct wlr_event_pointer_axis* event);
   void (*frame)(struct zn_cursor_grab* grab);
+  void (*cancel)(struct zn_cursor_grab* grab);
 };
 
 struct zn_cursor_grab {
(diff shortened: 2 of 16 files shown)


watasuke102:
Solve #147 at first

zen/scene/view.c
```diff
@@ -66,6 +70,19 @@ zn_view_get_window_fbox(struct zn_view *self, struct wlr_fbox *fbox)
...
+zn_view_change_board(struct zn_view *self, struct zn_board *new_board)
+{
+  zn_view_damage_whole(self);
+
+  wl_list_remove(&self->link);
+  wl_list_init(&self->link);
+  wl_list_insert(new_board->view_list.prev, &self->link);
+  self->board = new_board;
+
+  zn_view_damage_whole(self);
```
Aki-7, line 83: It's wired to call this here.
As changing the view's board without changing its position is not a common operation, this function doesn't seem visual operation. I mean no one changes the view's position only by using this function. So, damaging the view area after only changing the view's board doesn't make sense to me.

My suggestion is to create a function like,
```c
/** board can be NULL only when unmap */
void
zn_view_move(struct zn_view* self, struct zn_board* board, double x, double y) {
  zn_view_damage_whole(self);

  if (board != self->board) {
    if (self->board) { /* remove */ }
    if (board) { /* assign */}
  }
  
  self->x = x;
  self->y = y;

  if (self->impl->configure) {
    self->impl->configure(self, x, y);
  }

  zn_view_damage_whole(self);
}
```

zen/cursor.c
```diff
@@ -186,6 +203,7 @@ zn_cursor_handle_screen_destroy(struct wl_listener* listener, void* data)
     zn_cursor_update_position(self, screen, box.width / 2, box.height / 2);
     zn_cursor_damage_whole(self);
   } else {
+    self->grab->interface->cancel(self->grab);
```
Aki-7, line 206: Are you trying to make sure that non-default grabs are not used when there is no screen?

zen/input/cursor-grab-move.c
```diff
@@ -0,0 +1,111 @@
...
+  struct zn_cursor_grab_move* self =
+      zn_container_of(listener, self, view_unmap_listener);
+  zn_cursor_grab_move_end(self);
+}
+
+void
+zn_cursor_grab_move_end(struct zn_cursor_grab_move* self)
+{
+  wl_list_remove(&self->view_unmap_listener.link);
+  zn_cursor_set_xcursor(self->base.cursor, "left_ptr");
```
Aki-7, line 77: A xcursor is set also in `zn_cursor_end_grab`, which function, `zn_cursor_grab_move_end` or `zn_cursor_end_grab`, has a responsibility to set cursor image?

Aki-7 (changes requested):

zen/input/cursor-grab-move.c
```diff
@@ -0,0 +1,111 @@
...
+static void
+move_grab_motion(
+    struct zn_cursor_grab* grab, struct wlr_event_pointer_motion* event)
+{
+  UNUSED(event);
+  struct zn_cursor_grab_move* self = zn_container_of(grab, self, base);
+
+  if (grab->cursor->screen != self->prev_screen) {
+    zn_view_change_board(self->view, grab->cursor->screen->current_board);
+  }
```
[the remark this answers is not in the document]
watasuke102: Please read `How to check behavior`.

> drag between screens - the view is moved between boards

zen/cursor.c
watasuke102: yes

zen/cursor.c
```diff
@@ -221,6 +239,36 @@ zn_cursor_handle_surface_destroy(struct wl_listener* listener, void* data)
...
+  self->grab = &self->grab_default;
+
+  if (!cursor->screen) {
+    return;
+  }
+
+  surface = get_surface_of_view_at_cursor_pos(&view_x, &view_y);
+
+  if (surface) {
+    wlr_seat_pointer_enter(seat, surface, view_x, view_y);
```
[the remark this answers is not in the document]
watasuke102: What do you mean?

zen/input/cursor-grab-move.c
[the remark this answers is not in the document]
Aki-7: What about the case while switch board (using short cut) while moving?

zen/cursor.c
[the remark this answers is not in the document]
Aki-7: With the previous code, if the grab ends and the pointer enters a surface, the surface texture is not set.
If the client sends a "set cursor" request immediately after sending a "pointer enter" event, there is no problem, but if the client will not send a "set cursor" request, I though we should show a default cursor.

zen/cursor.c
Aki-7: That's a good very good constraint!
I would like you to add a comment about that.

zen/cursor.c
```diff
@@ -91,13 +85,37 @@ default_grab_frame(struct zn_cursor_grab* grab)
...
     .axis = default_grab_axis,
     .frame = default_grab_frame,
+    .cancel = default_grab_cancel,
 };
 
+// fetch view's surface corresponding to cursor's pos
+// expect cursor::screen is non-NULL
+static struct wlr_surface*
+zn_cursor_get_surface_at_cursor_pos(
+    struct zn_cursor* self, double* view_x, double* view_y)
```
Aki-7, line 106:
```suggestion
static struct wlr_surface*
zn_cursor_get_pointing_surface(
    struct zn_cursor* self, double* surface_x, double* surface_y)
```

include/zen/input/cursor-grab-move.h
```diff
@@ -0,0 +1,23 @@
...
+#include "zen/cursor.h"
+#include "zen/input/cursor-grab.h"
+#include "zen/scene/view.h"
+
+struct zn_cursor_grab_move {
+  struct zn_cursor_grab base;
+  struct zn_view* view;
+  struct zn_screen* prev_screen;
+
+  struct zn_board* init_board;
```
Aki-7, line 13: Please handle the destruction of the board.

I think we don't need to reset the position when canceled though.

include/zen/input/cursor-grab-move.h
```diff
@@ -0,0 +1,23 @@
+#ifndef ZN_CURSOR_GRAB_MOVE_H
+#define ZN_CURSOR_GRAB_MOVE_H
+
+#include "zen/cursor.h"
+#include "zen/input/cursor-grab.h"
+#include "zen/scene/view.h"
+
+struct zn_cursor_grab_move {
+  struct zn_cursor_grab base;
+  struct zn_view* view;
```
Aki-7, line 10:
```suggestion
  struct zn_view* view; // nonnull
```

zen/input/cursor-grab-move.c
```diff
@@ -0,0 +1,157 @@
...
+      zn_cursor_grab_move_handle_prev_screen_destroy;
+  wl_list_init(&self->prev_screen_destroy_listener.link);
+
+  return self;
+}
+
+static void
+zn_cursor_grab_move_destroy(struct zn_cursor_grab_move* self)
+{
+  wl_list_remove(&self->view_unmap_listener.link);
```
Aki-7, line 132:
```suggestion
  wl_list_remove(&self->view_unmap_listener.link);
  wl_list_remove(&self->prev_screen_destroy_listener.link);
```

zen/input/cursor-grab-move.c
```diff
@@ -0,0 +1,157 @@
...
+move_grab_motion(
+    struct zn_cursor_grab* grab, struct wlr_event_pointer_motion* event)
+{
+  UNUSED(event);
+  struct zn_cursor_grab_move* self = zn_container_of(grab, self, base);
+  struct zn_board* board = self->view->board;
+
+  if (!grab->cursor->screen) {
+    return;
+  }
```
Aki-7, line 17: With this constraint: while a non-default grab is in use, the cursor's screen will not be NULL, `cursor->screen` must not be NULL here.

zen/scene/view.c
```diff
@@ -9,15 +9,32 @@
...
+    struct zn_view *self, double x, double y, struct zn_board *new_board)
 {
+  zn_view_damage_whole(self);
+
+  if (new_board != self->board) {
+    if (self->board) {
+      wl_list_remove(&self->link);
+      wl_list_init(&self->link);
+    }
+    // new_board is NULL only when unmapping
```
Aki-7, line 23: This comment should be at the prototype declaration.
(I added another review  about this)

zen/scene/view.c
```diff
@@ -93,8 +110,8 @@ zn_view_map_to_scene(struct zn_view *self, struct zn_scene *scene)
   // TODO: handle board destruction
 
   zn_view_get_window_fbox(self, &fbox);
-  zn_view_move(
-      self, (board->width - fbox.width) / 2, (board->height - fbox.height) / 2);
+  zn_view_move(self, (board->width - fbox.width) / 2,
+      (board->height - fbox.height) / 2, self->board);
 
   self->board = board;
   wl_list_insert((&board->view_list)->prev, &self->link);
```
Aki-7, line 117:
```suggestion
  zn_view_move(self, (board->width - fbox.width) / 2,
      (board->height - fbox.height) / 2, board);
```

Aki-7 (changes requested):

include/zen/scene/view.h
```diff
@@ -38,6 +38,9 @@ struct zn_view {
   } events;
 };
 
+void zn_view_move(
+    struct zn_view *self, double x, double y, struct zn_board *new_board);
```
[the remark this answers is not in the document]
watasuke102: Originally, I think `zn_view_move` only updates the position of the view, so updating board is secondaly task. Arguments about position seem more important for me, and my prefer important argument should place first, so I added argument of the board to the last.

include/zen/input/cursor-grab.h
```diff
@@ -17,6 +17,8 @@ struct zn_cursor_grab_interface {
   void (*axis)(
       struct zn_cursor_grab* grab, struct wlr_event_pointer_axis* event);
   void (*frame)(struct zn_cursor_grab* grab);
+  // called when zn_cursor::screen is assigned NULL
```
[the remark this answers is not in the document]
watasuke102: > zn_cursor_grab::cancel is the interface to cancel the current grab operation, not the interface which is called when zn_cursor::screen is assigned NULL.

That's true, but I believe it is also correct to call `zn_cursor_grab::cancel` when `zn_cursor::screen` is assigned NULL. I mean with this comment is that this interface will be called when screen is assigned NULL, not that this is the only case.

And, I wonder that problems about screen's constraint is solved by calling `zn_cursor_grab::cancel` when `zn_cursor::screen` is assigned NULL.

zen/input/cursor-grab-move.c
```diff
@@ -0,0 +1,157 @@
...
+move_grab_frame(struct zn_cursor_grab* grab)
+{
+  UNUSED(grab);
+}
+
+static void
+move_grab_rebase(struct zn_cursor_grab* grab)
+{
+  // the board which the view is located is changed
+  move_grab_cancel(grab);
```
Aki-7, line 56: In my opinion, If boards are switched while a view is being moved, it is better to move the view to the board that is displayed and continue the move operation.

```suggestion
  struct zn_cursor_grab_move* self = zn_container_of(grab, self, base);
  struct zn_board* board =
      zn_screen_get_current_board(self->base.cursor->screen);
  zn_view_move(self->view, board, self->base.cursor->x + self->diff_x,
      self->base.cursor->y + self->diff_y);
```

Aki-7 (changes requested):

zen/input/cursor-grab-move.c
```diff
@@ -0,0 +1,157 @@
...
+static void
+move_grab_frame(struct zn_cursor_grab* grab)
+{
+  UNUSED(grab);
+}
+
+static void
+move_grab_rebase(struct zn_cursor_grab* grab)
+{
+  // the board which the view is located is changed
```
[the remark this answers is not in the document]
watasuke102: > Not limited to when the board on which the view is located is changed

What is the case for example?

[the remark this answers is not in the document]
Aki-7: An example would be when a new view is created during the move operation.
You may want to raise the moving view above the newly generated view.

But my point has nothing to do with what the example is. The point is to clarify the responsibility of this `zn_cursor_grab::rebase` interface.
Since we are providing polymorphism by introducing a grab, the rebase of default_grab and the rebase of move_grab must handle the same situation. That situation is, as the name `rebase` suggests, repositioning the cursor.
What kind of situation is meant by repositioning the cursor is a matter of debate, and I hope that a common understanding will emerge in the course of development, but at least I don't think it refers only to switching boards.
In other words, it is the role of screens to handle board switching, and the function rebase of zn_cursor should be used to realize board switching, and zn_cursor itself should not know as much as possible that the board has switched, etc.

Let me explain from a different perspective. But I believe this is just a more specific example of what I mentioned above.
If we make the rebase of move_grab only for when switching boards, that means that it would be a mistake to call `cursor->grab->interface->rabase()` at any other time even for default garb. Or you need to make sure that `cursor->grab->interface` matches the interface of move_grab before calling it. This would violate the concept of polymorphism.

Some may argue that using polymorphism is a bad idea. That is certainly a possibility. But we are able to describe cursor behavior relatively safely with it, which is probably why other windowing systems also use this polymorphism.

Aki-7 (approved):
That's a lot of work, great job!

Aki-7 (approved):

PR #146, opened 2022-09-04
insert new views to the tail of the view_list

liwii:
## Context
#125
## Summary
When a new view is mapped to the scene with `zn_view_map_to_scene`, insert the view to the tail of the `view_list` so that the newest view is shown on the top of the view stack.
## How to check behavior
Run
```
zen-desktop -c weston-terminal
WAYLAND_DISPLAY=wayland-1 weston-flower
```
and confirm that the weston-flower client is shown on top of the terrminal.


Aki-7 (approved):
LGTM!

Issue #147, opened 2022-09-06
view::x, y is different from the top-left of the window


watasuke102:
Now, move the view to (0,0), view is rendered the following:

![image](https://user-images.githubusercontent.com/36789813/188534724-934a51f7-622a-4749-bad1-f66143d34fa5.png)

The margin like this is not expected.

PR #148, opened 2022-09-06
fix assigning view's position process

watasuke102:
## Context

See #147

## Summary

Use view's geometry.

## How to check behavior

Move the view to any position, and check that there is no difference between specified pos and appeared pos. 

try: edit `view.c@zn_view_map_to_scene` like following:
```diff
  zn_view_move(
-      self, (board->width - fbox.width) / 2, (board->height - fbox.height) / 2);
+      self, 0, 0);
       // or
+      self, (board->width - fbox.width), (board->height - fbox.height));
```

### DIFF
--- a/zen/scene/view.c
+++ b/zen/scene/view.c
@@ -12,8 +12,13 @@
 static void
 zn_view_move(struct zn_view *self, double x, double y)
 {
-  self->x = x;
-  self->y = y;
+  struct wlr_box view_geometry;
+
+  zn_view_damage_whole(self);
+
+  self->impl->get_geometry(self, &view_geometry);
+  self->x = x - view_geometry.x;
+  self->y = y - view_geometry.y;
 
   if (self->impl->configure) {
     self->impl->configure(self, x, y);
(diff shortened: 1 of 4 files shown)


zen/scene/view.c
```diff
@@ -12,8 +12,13 @@
...
 {
-  self->x = x;
-  self->y = y;
+  struct wlr_box view_geometry;
+
+  zn_view_damage_whole(self);
+
+  self->impl->get_geometry(self, &view_geometry);
+  self->x = x - view_geometry.x;
+  self->y = y - view_geometry.y;
```
Aki-7, line 21: It's a little bit confusing that the x and y of `zn_view_move(struct zn_view *self, double x, double y)` is in the different coordinate system from that of `zn_view::x` and `zn_view::y`.
How about redefining the coordinate system of `zn_view::x` and `zn_view::y`, such that the origin of the coordinate system is at the left top of the view geometry?

watasuke102: They are renamed `surface_{x,y}`, and arguments of `zn_view_move()` are also renamed `window_{x,y}`

PR #149, opened 2022-09-06
fix header include

Aki-7:
## Context

DRM headers seem to be located in different paths depending on the developers' environment.
I, at least, confirmed two locations.
- /usr/include/libdrm
- /usr/include/drm

`pkgconfig` absorb the difference by setting the Cflags `-I/usr/include/libdrm` or `-I/usr/include/drm`, so we should not do
```c
#include <libdrm/drm.h>
```

instead, do
```c
#include <drm.h>
```

## Summary

- [x] fix header include

## How to check behavior

No change

### DIFF
--- a/zen/cursor.c
+++ b/zen/cursor.c
@@ -1,6 +1,6 @@
 #include "zen/cursor.h"
 
-#include <drm/drm_fourcc.h>
+#include <drm_fourcc.h>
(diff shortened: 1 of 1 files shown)


watasuke102 (approved):
✨**LGTM**✨

PR #155, opened 2022-09-12
restack view

watasuke102:
## Context

Currently, the view is focused when clicked, but its order isn't changed.

## Summary

Change the order of the view when the view is clicked.

## How to check behavior

1. start zen with more than two clients
2. click the view one by one

### DIFF
--- a/zen/scene/view.c
+++ b/zen/scene/view.c
@@ -10,6 +10,19 @@
 #include "zen/xwayland-view.h"
 
 void
+zn_view_bring_to_front(struct zn_view *self)
+{
+  wl_list_remove(&self->link);
+  wl_list_insert(self->board->view_list.prev, &self->link);
+
+  if (self->impl->restack) {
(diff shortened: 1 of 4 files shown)


zen/scene/view.c
```diff
@@ -151,6 +164,10 @@ zn_view_unmap(struct zn_view *self)
 {
   wl_signal_emit(&self->events.unmap, NULL);
 
+  if (self->impl->restack) {
+    self->impl->restack(self, XCB_STACK_MODE_BELOW);
+  }
+
```
Aki-7, line 170: [Question]
Why do we need to send this view to back?
I though this view will no longer be in the stack soon.

watasuke102: Sorry, this change is little different from this PR's intend, it will fix another issue (not exist on GitHub).

1. start zen
2. create board
3. start x client on each boards
4. cursor events are not sent to visible views which is launched later

PR #156, opened 2022-09-13
resize view

watasuke102:
## Context

Resize the view by the cursor.

## Summary

Add `zn_cursor_grab_resize`.

## How to check behavior

1. start zen and some wayland client
2. drag the corner of the view
3. the size of the view is changed following the cursor

### DIFF
--- a/include/zen/input/cursor-grab-resize.h
+++ b/include/zen/input/cursor-grab-resize.h
@@ -7,7 +7,6 @@
 struct zn_cursor_grab_resize {
   struct zn_cursor_grab base;
   struct zn_view* view;
-  uint32_t edges;
 
   double init_view_x, init_view_y;
   double init_view_width, init_view_height;
--- a/include/zen/input/cursor-grab-resize.h
+++ b/include/zen/input/cursor-grab-resize.h
@@ -1,5 +1,4 @@
-#ifndef ZN_CURSOR_GRAB_RESIZE_H
-#define ZN_CURSOR_GRAB_RESIZE_H
+#pragma once
 
 #include "zen/cursor.h"
 #include "zen/input/cursor-grab.h"
@@ -20,5 +19,3 @@ struct zn_cursor_grab_resize {
 
 void zn_cursor_grab_resize_start(
     struct zn_cursor* cursor, struct zn_view* view, uint32_t edges);
-
-#endif  //  ZN_CURSOR_GRAB_RESIZE_H
--- a/include/zen/input/cursor-grab-resize.h
+++ b/include/zen/input/cursor-grab-resize.h
@@ -9,10 +9,10 @@ struct zn_cursor_grab_resize {
   struct zn_view* view;
   uint32_t edges;
 
-  struct wlr_fbox init_view_box;
+  double init_view_x, init_view_y;
+  double init_view_width, init_view_height;
+
   double init_cursor_x, init_cursor_y;
-  // distance between cursor's pos and window's pos
-  double diff_x, diff_y;
 
   struct wl_listener view_unmap_listener;
 };
--- a/include/zen/input/cursor-grab-resize.h
+++ b/include/zen/input/cursor-grab-resize.h
@@ -8,8 +8,12 @@
 struct zn_cursor_grab_resize {
   struct zn_cursor_grab base;
   struct zn_view* view;
-  struct wlr_fbox init_box;
   uint32_t edges;
+
+  struct wlr_fbox init_view_box;
+  double init_cursor_x, init_cursor_y;
+  // distance between cursor's pos and window's pos
+  double diff_x, diff_y;
 
   struct wl_listener view_unmap_listener;
 };
--- a/include/zen/input/cursor-grab-resize.h
+++ b/include/zen/input/cursor-grab-resize.h
@@ -1,0 +1,20 @@
+#ifndef ZN_CURSOR_GRAB_RESIZE_H
+#define ZN_CURSOR_GRAB_RESIZE_H
+
+#include "zen/cursor.h"
+#include "zen/input/cursor-grab.h"
+#include "zen/scene/view.h"
+
+struct zn_cursor_grab_resize {
+  struct zn_cursor_grab base;
+  struct zn_view* view;
+  struct wlr_fbox init_box;
+  uint32_t edges;
+
+  struct wl_listener view_unmap_listener;
+};
+
+void zn_cursor_grab_resize_start(
+    struct zn_cursor* cursor, struct zn_view* view, uint32_t edges);
+
+#endif  //  ZN_CURSOR_GRAB_RESIZE_H
--- a/include/zen/scene/view.h
+++ b/include/zen/scene/view.h
@@ -44,11 +44,10 @@ struct zn_view {
 
 void zn_view_bring_to_front(struct zn_view *self);
 
-// set position and size by fbox
 // box::x, y          -> surface's pos (corresponding to view::x, y)
 // box::width, height -> window's size
 //                      (corresponding to the result of view_get_window_fbox)
-void zn_view_configure_by_fbox(struct zn_view *self, struct wlr_fbox *box);
+void zn_view_configure(struct zn_view *self, struct wlr_fbox *box);
 
 /**
  * @param board must not be NULL except when this view is unmapped with
--- a/include/zen/scene/view.h
+++ b/include/zen/scene/view.h
@@ -38,7 +38,7 @@ struct zn_view {
   struct zn_board *board;  // non null, when mapped
 
   struct {
-    bool resizing;
+    enum { None, Resizing, Canceled } mode;
     uint32_t edges;
   } resize_status;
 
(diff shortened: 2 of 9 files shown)


include/zen/input/cursor-grab-resize.h
```diff
@@ -0,0 +1,24 @@
+#ifndef ZN_CURSOR_GRAB_RESIZE_H
```
Aki-7, line 1: Please use `#pragma once`

include/zen/scene/view.h
```diff
@@ -35,17 +36,30 @@ struct zn_view {
...
+/** set window's size (corresponding to the result of view_get_window_fbox)
+ * @retval serial for resize event
+ */
+uint32_t zn_view_set_size(struct zn_view *self, double width, double height);
+
 /**
  * @param board must not be NULL except when this view is unmapped with
  * `zn_view_unmap`
  */
+
```
Aki-7, line 62:
```suggestion
```

include/zen/scene/view.h
```diff
@@ -35,17 +36,30 @@ struct zn_view {
...
   struct {
     struct wl_signal unmap;
   } events;
 };
 
 void zn_view_bring_to_front(struct zn_view *self);
 
+/** set window's size (corresponding to the result of view_get_window_fbox)
+ * @retval serial for resize event
+ */
```
Aki-7, line 55:
```suggestion
/** 
 * set window's size (corresponding to the result of view_get_window_fbox)
 * @retval serial for resize event
 */
```

zen/scene/view.c
```diff
@@ -22,6 +22,14 @@ zn_view_bring_to_front(struct zn_view *self)
   zn_view_damage_whole(self);
 }
 
+uint32_t
+zn_view_set_size(struct zn_view *self, double width, double height)
+{
+  zn_view_damage_whole(self);
```
Aki-7, line 28: This seems unnecessary.

include/zen/scene/view.h
```diff
@@ -12,9 +12,10 @@ struct zn_view;
 struct zn_view_impl {
   struct wlr_surface *(*get_wlr_surface)(struct zn_view *view);
   void (*get_geometry)(struct zn_view *view, struct wlr_box *box);
-  void (*configure)(struct zn_view *view, double x, double y);  // nullable
   void (*for_each_popup_surface)(struct zn_view *view,
       wlr_surface_iterator_func_t iterator, void *user_data);  // nullable
+  uint32_t (*set_size)(struct zn_view *view, double width, double height);
+  void (*set_position)(struct zn_view *view, double x, double y);  // nulable
```
Aki-7, line 18: I feel it is not a good idea to separate `configure` into `set_size` and `set_position`.
I think that when you have the resizing process revised in this PR, you will need to atomically change the size and position of the view.

zen/xdg-toplevel-view.c
```diff
@@ -71,6 +72,48 @@ zn_xdg_toplevel_view_move_handler(struct wl_listener* listener, void* data)
...
+  }
+
+  double dx = self->base.resize_status.requested_box.width -
+              event->surface->pending.geometry.width;
+  double dy = self->base.resize_status.requested_box.height -
+              event->surface->pending.geometry.height;
+
+  zn_view_move(&self->base, self->base.board,
+      self->base.resize_status.requested_box.x + dx,
+      self->base.resize_status.requested_box.y + dy);
```
Aki-7, line 114: The actual change of view position should be at the time of `commit`. Until `commit`, the surface state is pending and should not be reflected in the rendering.

zen/xdg-toplevel-view.c
```diff
@@ -156,6 +209,15 @@ zn_xdg_toplevel_view_create(
   self->move_listener.notify = zn_xdg_toplevel_view_move_handler;
   wl_list_init(&self->move_listener.link);
 
+  self->resize_listener.notify = zn_xdg_toplevel_view_resize_handler;
+  wl_signal_add(
+      &self->wlr_xdg_toplevel->events.request_resize, &self->resize_listener);
```
Aki-7, line 214: Please register this listener on `map`.

Aki-7 (changes requested):
1. I think the calculation of view size and coordinates when resizing is too complicated. Please consider if it can be refactored.

2. Please rebase to main branch. I don't trust git auto merging / rebasing.

include/zen/scene/view.h
watasuke102: I don't think so.
xdg_toplevel doesn't provide the function of updating its position, so they don't seem the same. Also, you reported before that the size of `weston-terminal` is changed on moving the view, it may occured because `configure` also updates the **size**, despite unnecessary. Separating these will solve this problem.

watasuke102:
> Please consider if it can be refactored

I can't

include/zen/xwayland-view.h
```diff
@@ -15,6 +15,7 @@ struct zn_xwayland_view {
   struct wl_listener map_listener;
   struct wl_listener unmap_listener;
   struct wl_listener move_listener;
+  struct wl_listener resize_listener;
```
Aki-7, line 18: Will not implement actual resizing for xwayland-view?

zen/input/cursor-grab-resize.c
```diff
@@ -0,0 +1,182 @@
...
+  UNUSED(grab);
+}
+
+static void
+resize_grab_rebase(struct zn_cursor_grab* grab)
+{
+  struct zn_cursor_grab_resize* self = zn_container_of(grab, self, base);
+  struct zn_board* board = zn_screen_get_current_board(grab->cursor->screen);
+  zn_view_move(self->view, board, grab->cursor->x, grab->cursor->y);
+}
```
Aki-7, line 72: In the case of resizing, it would be more natural to cancel at rebasing. You can ask other developers about this.

zen/input/cursor-grab-resize.c
```diff
@@ -0,0 +1,182 @@
...
+
+static void
+resize_grab_cancel(struct zn_cursor_grab* grab)
+{
+  struct zn_cursor_grab_resize* self = zn_container_of(grab, self, base);
+  zn_view_move(
+      self->view, self->view->board, self->init_view_x, self->init_view_y);
```
Aki-7, line 79: This seems needless because `set_size` will trigger the next commit with the resized surface, and then, at the time of commit, the view's position will be moved.

zen/input/cursor-grab-resize.c
```diff
@@ -0,0 +1,182 @@
...
+static void
+zn_cursor_grab_resize_destroy(struct zn_cursor_grab_resize* self)
+{
+  wl_list_remove(&self->view_unmap_listener.link);
+  free(self);
+}
+
+static void
+zn_cursor_grab_resize_end(struct zn_cursor_grab_resize* self)
+{
```
Aki-7, line 148:
```suggestion
{
  zn_cursor_set_xcursor(self->base.cursor, "left_ptr");
```

Aki-7 (changes requested):
Nice refactoring!

include/zen/xwayland-view.h
watasuke102: I eager to implement this, but I don't have any way to test it (x client is't able to resize now). If it is allowed, I'll leave FIXME comment.

Aki-7: I see,
> I'll leave FIXME comment.

It's good enough!

zen/xdg-toplevel-view.c
```diff
@@ -18,19 +18,22 @@ zn_xdg_toplevel_view_wlr_surface_commit_handler(
...
+  if (self->base.resize_edges & WLR_EDGE_TOP) {
+    dy = surface->previous.height - surface->current.height;
   }
 
-  struct wlr_surface* surface = data;
-  const int dx = surface->previous.width - surface->current.width;
-  const int dy = surface->previous.height - surface->current.height;
   zn_view_move(
       &self->base, self->base.board, self->base.x + dx, self->base.y + dy);
+  self->base.resize_edges = 0;
```
Aki-7, line 36: I think `resize_edges` should not be reset (and should not set a value in `resize_grab_motion`).

Consider the case where the events(→) and requests(←) happen in the following order,
→ configure ( = set_size)
→ configure ( = set_size)
← commit (using the data of the first configure event)
← commit (using the data of the second configure event)

the last commit will not be treated properly.

Aki-7 (changes requested):

include/zen/scene/view.h
```diff
@@ -37,8 +37,10 @@ struct zn_view {
   struct wl_list link;     // zn_board::view_list;
   struct zn_board *board;  // non null, when mapped
 
-  // when this isn't 0, the view is resized by grab
-  uint32_t resize_edges;
+  struct {
+    enum { None, Resizing, Canceled } mode;
```
Aki-7, line 41: This still does not resolve the case I mentioned in https://github.com/zigen-project/zen/pull/156#discussion_r973874231
The solution is more simple, just previous `bool resizing` seems good to me.

Aki-7: And this enum can cause namespace conflict. Please use the UPPER_CASE, global-unique name for enum members.

watasuke102: > still does not resolve

What is the case you thinking? Unless sending two or more configure event, the mode isn't changed, so I thought it works well.

> bool resizing seems good

How can it process canceling (update position)? On previous code, `resizing` is set false when canceling.

Aki-7 (approved):
Looks Good To Me!

Aki-7 (approved):

Issue #123, opened 2022-08-22
Enable to change output's scale and transform


Aki-7:
## Current status

We are trying to support output scaling and rotation, but no means to set those properties.

## Expected

Enable to set those properties somehow.

PR #160, opened 2022-09-19
Configure bg images with .config/zen-desktop/config.toml

liwii:
## Context
To realize user-specific configuration like #123, we need a way to give configuration data to zen-desktop after the build.

## Summary
Let users specify their background image with a `toml` config file.Following [XDG Base Directory Specification
](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), it first searches `$XDG_CONFIG_HOME/zen-desktop/config.toml`, then falls back to `$HOME/.config/zen-desktop/config.toml`.

## How to check behavior
- Create the following config file and save it as `$HOME/.config/zen-desktop/config.toml`. Note that only **PNG images with RGBA format** can be used as the background images for now.
```toml
[background]
image = "/path/to/your/image/file"
```
- Run `zen-desktop` and see if the change is reflected.


Aki-7:
TIPS (for others who try this with their favorite image)

ImageMagick command
```
convert image.jpeg -define png:color-type=6 image.png
```

zen/config.c
```diff
@@ -0,0 +1,77 @@
...
+    toml_datum_t image = toml_string_in(bg, "image");
+    self->bg_image_file = image.u.s;
+  }
+  toml_free(tbl);
+  return self;
+err:
+  return NULL;
+}
+void
+zn_config_destroy(struct zn_config *self)
```
watasuke102, line 74: (nit) please add blank line.

```suggestion
}

void
zn_config_destroy(struct zn_config *self)
```

watasuke102 (changes requested):

include/zen/config.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+struct zn_config {
+  char* bg_image_file;
+};
+
+struct zn_config* zn_config_create(void);
+void zn_config_destroy(struct zn_config* self);
```
watasuke102, line 8: and here
```suggestion
void zn_config_destroy(struct zn_config* self);

```

zen/server.c
```diff
@@ -338,6 +348,7 @@ zn_server_create(struct wl_display *display)
 void
 zn_server_destroy(struct zn_server *self)
 {
+  zn_config_destroy(self->config);
```
Aki-7, line 351: Basically, it is better to delete objects in the reverse order of their creation.
In this case, `zc_config_destroy` should be called after `sever_singleton = NULL`.

Aki-7 (changes requested):
As you know, string (char*) allocates a certain contiguous part of memory region. For this reason, we need to be very careful when manipulating strings.

include/zen/config.h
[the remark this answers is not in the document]
liwii: Thanks,
I think the editor is following `.clang-format` in the project. Is there any good config to do this eof new line thing?

include/zen/config.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+struct zn_config {
+  char* bg_image_file;
```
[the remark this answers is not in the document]
liwii: What's the intention of `const`? If it is eventually freed, it does not make sense to mark it as `const` (at least I feel a bit uncomfortable to do const to non-const casting in `free`.

include/zen/config.h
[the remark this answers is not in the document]
Aki-7: You are right, clang-format should handle this issue. 

>  Is there any good config to do this eof new line thing?

I don't know. Could you add a new config for this if any.

zen/config.c
```diff
@@ -0,0 +1,77 @@
...
+  char *config_home;
+  config_home = getenv("XDG_CONFIG_HOME");
+  if (config_home == NULL) {
+    char *homedir = getenv("HOME");
+    if (homedir == NULL) homedir = getpwuid(getuid())->pw_dir;
+    if (homedir == NULL) {
+      zn_warn("Could not find the home directory");
+      return NULL;
+    }
+    config_home = strcat(homedir, "/.config");
```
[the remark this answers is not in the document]
liwii: Thanks for pointing that out. I didn't know that the directly manipulating the pointer obtained with `getenv()` is that dangerous.

include/zen/config.h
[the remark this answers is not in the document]
Aki-7: Sorry, this is my mistake. I added "const" because DEFAULT_WALLPAPER was assigned to it, but as you said, we should not.

zen/config.c
```diff
@@ -13,23 +13,35 @@
...
+    strcpy(config_home, homedir_env);
+    strcat(config_home, config_dir_partial_path);
+  } else {
+    config_home = (char *)malloc(
+        strlen(config_home_env) + strlen(toml_file_partial_path) + 1);
+    strcpy(config_home, config_home_env);
   }
 
-  char *config_filename = strcat(config_home, "/zen-desktop/config.toml");
+  strcat(config_home, toml_file_partial_path);
```
Aki-7, line 40: Instead of `strcat` and `strcpy`, how about using `sprintf`?
This is a matter of preference, so if you don't like it, ignore it.
```c
if () {
   config_home = malloc(...);
   sprintf(config_home, "%s%s%s" ...);
else {
   config_home = malloc(...);
   sprintf(config_home, "%s%s" ...);
}
```

liwii: `sprintf` looks much cleaner. I fixed the code to use it. Thanks.

zen/config.c
```diff
@@ -0,0 +1,99 @@
...
+        strlen(config_home_env) + strlen(toml_file_partial_path) + 1);
+    sprintf(config_home, "%s%s", config_home_env, toml_file_partial_path);
+  }
+
+  FILE *fp = fopen(config_home, "r");
+  free(config_home);
+  if (fp == NULL) {
+    zn_warn("config.toml not found in %s\n", config_home);
+    return NULL;
+  }
```
watasuke102, line 44: Accessing the released resource (`config_home`)

zen/config.c
```diff
@@ -0,0 +1,99 @@
...
+  if (!tbl) {
+    zn_warn("Could not get the toml table");
+    return self;
+  }
+
+  toml_table_t *bg = toml_table_in(tbl, "background");
+  if (bg != NULL) {
+    toml_datum_t image = toml_string_in(bg, "image");
+    self->bg_image_file = image.u.s;
+  }
```
watasuke102, line 85: At this timing, `self->bg_image_file` has duplicated `DEFAULT_WALLPAPER` (by `zn_config_set_default`), so calling `free()` is required.

watasuke102 (changes requested):

zen/config.c
```diff
@@ -0,0 +1,99 @@
...
+  } else {
+    config_home = (char *)malloc(
+        strlen(config_home_env) + strlen(toml_file_partial_path) + 1);
+    sprintf(config_home, "%s%s", config_home_env, toml_file_partial_path);
+  }
+
+  FILE *fp = fopen(config_home, "r");
+  free(config_home);
+  if (fp == NULL) {
+    zn_warn("config.toml not found in %s\n", config_home);
```
Aki-7, line 42: It could be for reasons other than the file not existing.

```suggestion
    zn_warn("Failed to open %s\n", config_home);
```

Aki-7 (commented):
I left a few minor comments, but basically I think it is good except for the part of watasuke's comment.

zen/config.c
```diff
@@ -0,0 +1,99 @@
...
+  if (self == NULL) {
+    zn_error("Failed to allocate memory");
+    goto err;
+  }
+
+  zn_config_set_default(self);
+
+  toml_table_t *tbl = zn_config_get_toml_table();
+  if (!tbl) {
+    zn_warn("Could not get the toml table");
```
Aki-7, line 77: I think it's better to use words like "Fail" and "Error" as much as possible, because they are easy to search.
```suggestion
    zn_warn("Failed to get the toml table");
```

watasuke102 (approved):
**_LGTM_**

Aki-7 (approved):
Looks good to me!

watasuke102 (dismissed):

watasuke102 (approved):

Aki-7 (approved):

PR #161, opened 2022-09-19
Send keyboard events

watasuke102:
## Context

Send keyboard events to client

## Summary

focus seat_keyboard to the view's surface

## How to check behavior

1. start zen with any client (reccomend: `weston-terminal`)
2. press any key (try also Tab or arrow key)
3. keyboard events will be sent

### DIFF
--- a/zen/server.c
+++ b/zen/server.c
@@ -336,7 +336,6 @@ err:
 void
 zn_server_destroy(struct zn_server *self)
 {
-  zn_scene_set_focused_view(self->scene, NULL);
   wlr_xwayland_destroy(self->xwayland);
   zn_input_manager_destroy(self->input_manager);
   zn_display_system_destroy(self->display_system);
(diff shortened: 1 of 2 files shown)


zen/server.c
```diff
@@ -336,6 +336,7 @@ zn_server_create(struct wl_display *display)
 void
 zn_server_destroy(struct zn_server *self)
 {
+  zn_scene_set_focused_view(self->scene, NULL);
```
watasuke102, line 339: This is preventing Segmentation Fault, but it seems not good way. Do you have any idea?

watasuke102: ?

```suggestion
  wlr_xwayland_destroy(self->xwayland);
  wl_display_destroy_clients(self->display);
```

watasuke102: I'll ignore this SEGV (but leave the issue, and solve another PR)

Aki-7 (approved):
Great!

PR #163, opened 2022-09-20
add key binding for switching virtual terminal

watasuke102:
## Context

At other DE, when press Ctrl+Alt+F1~, it switches to virtual terminal (tty).

## Summary

Implement switching vt by `wlr_session_change_vt`

## How to check behavior

1. start zen
2. press Ctrl+Alt+F([1-9]|10)
3. switch to virtual terminal

### DIFF
--- a/zen/scene/scene.c
+++ b/zen/scene/scene.c
@@ -3,6 +3,8 @@
 #include <cairo.h>
 #include <drm_fourcc.h>
 #include <linux/input.h>
+#include <wlr/backend/drm.h>
+#include <wlr/backend/multi.h>
 
 #include "zen-common.h"
 #include "zen-config.h"
@@ -69,6 +71,22 @@ zn_scene_new_board_binding_handler(uint32_t time_msec, uint32_t key, void* data)
 
   zn_board_assign_to_screen(board, screen);
   zn_screen_set_current_board(screen, board);
+}
+
+static void
+zn_scene_switch_vt_handler(uint32_t time_msec, uint32_t key, void* data)
+{
+  UNUSED(data);
+  UNUSED(time_msec);
+  const unsigned int vt = key - KEY_F1 + 1;
+  struct zn_server* server = zn_server_get_singleton();
+  struct wlr_session* session = wlr_backend_get_session(server->backend);
+
+  if (!session) {
+    return;
+  }
+
+  wlr_session_change_vt(session, vt);
 }
 
 void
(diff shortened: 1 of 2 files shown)


zen/scene/scene.c
```diff
@@ -71,6 +73,22 @@ zn_scene_new_board_binding_handler(uint32_t time_msec, uint32_t key, void* data)
   zn_screen_set_current_board(screen, board);
 }
 
+static void
+zn_scene_switch_vt_handler(uint32_t time_msec, uint32_t key, void* data)
+{
+  UNUSED(data);
+  UNUSED(time_msec);
+  const unsigned int vt = key - KEY_F1 + 1;
```
shierote, line 81: do you need handling of cases where vt is less than 1?

shierote (approved):

zen/scene/scene.c
watasuke102: > vt is less than 1

What is the case? `zn_scene_switch_vt_handler` is registered to the range of F1-F10, so the range of values that `vt` takes is 1-10.

shierote: No case, but I asked just in case the argument `key` of this function is not necessarily `KEY_F1`~`KEY_F10`.

shierote: But of course, if the user calling this function can recall from the function name the contract that there are no other keys to pass to this function than `KEY_F1`~`KEY_F10`, then the case handle is not necessary.

shierote (dismissed):
nice change!

Aki-7 (changes requested):
I don't think switching virtual terminal is the responsibility of `scene`, but of `main.c`.
It's like handling signals.

zen/scene/scene.c
```diff
@@ -186,6 +210,14 @@ zn_scene_setup_bindings(struct zn_scene* self)
   zn_input_manager_add_key_binding(server->input_manager, KEY_N,
       WLR_MODIFIER_LOGO | WLR_MODIFIER_SHIFT,
       zn_scene_new_board_binding_handler, self);
+
+  if (wlr_backend_is_multi(server->backend)) {
```
Aki-7, line 214: At the conclusion of the last discussion, wasn't it sufficient to check for the existence of `session` in the handler, rather than checking for `is_multi` here?

watasuke102: Yes, it seems enough. But there is no reason to setup useless keybind, I think.

Aki-7: I know what you feel. But at the same time, this handles useless (never happen) situation, that is the backend is not the multi backend. It seems more useless for me. This causes the developer to read the code to see if the backend may not be multi-backend, as we did a few days ago.

zen/scene/scene.c
```diff
@@ -3,6 +3,8 @@
 #include <cairo.h>
 #include <drm_fourcc.h>
 #include <linux/input.h>
+#include <wlr/backend/drm.h>
+#include <wlr/backend/multi.h>
```
Aki-7, line 7:
```suggestion
```

Aki-7 (commented):
LGTM!

Aki-7 (approved):

Issue #164, opened 2022-09-20
not handling selection events: no seat assigned to xwayland


shierote:
`[wlr] [xwayland/selection/selection.c:147] not handling selection events: no seat assigned to xwayland`

Error that appears when using google-chrome or vscode's x-client. It seems that x-client popup cannot be handled well without fixing this.

Issue #116, opened 2022-08-21
support for xdg popup / xwayland popup / subsurface


shierote:
# xdg popup

- [x] use popup surface in effective coordinate system and transform to output/transformed coordinate/resolution. (this may be solved in https://github.com/zigen-project/zen/pull/121)
- [x] show surface (damage) #157 
- [x] hide surface (damage) #157 
- [x] input #165
  - receive the input events of wl_pointer, wl_keyboard, etc.
- [ ] surface damage while moving (view-child damage)
  - maybe some clients create popup while moving
- [ ] popups of popup (popup tree)
- [ ] investigate whether a NULL check of the parent view is required
- [ ] adjust popup's position; in case the popup is going to be outside the monitor.

# xwayland popup
 - maybe the surface whose override_redirect == true (check unmanaged view in sway.)
- [ ] #164 

# subsurface
- [ ] show surface (damage)
- [ ] hide surface (damage)
- [ ] surface damage while moving (view-child damage)
- [ ] input
  - receive the input events of wl_pointer, wl_keyboard, etc.
- [ ] subsurface tree
- [ ] adjust subusrface's position; in case the subsurface is going to be outside the monitor.

cf.
https://github.com/wayland-project/wayland-protocols/blob/main/stable/xdg-shell/xdg-shell.xml#L1140

PR #165, opened 2022-09-20
enable to input xdg popup

shierote:
## Context

#116 ( before: #157 )

## Summary

- return popup's surface by new `zn_screen_get_surface_at` function

## How to check behavior

The following behaviors can be checked with the Wayland clients (e.g., `vlc` `weston-terminal` `weston-flower`)

- When hovering over a menu item in the wayland client's popup menu, the color of the item changes.
  - **In particular, it is possible to hover a popup outside the frame of the view.**

- When you click (button event) on each menu of the wayland client's popup, the input works accordingly (except unsupported functions such as minimize/maximize).
  - **In particular, it is possible to click a popup outside the frame of the view.**

- After hiding a popup, input is sent to the view where the popup was (input is not sent to the popup).

- "hiding popup" should occur when...
  - (right/left) click elsewhere in the view
  - close the view
  - start `move`
  - focus/restack another view

Belows are TODO.
- popups of popup. (popup tree)
- the verification of xwayland popups will be done in another PR. (#164 may be related)

### DIFF
--- a/zen/scene/screen.c
+++ b/zen/scene/screen.c
@@ -6,7 +6,6 @@
 #include "zen/scene/screen-layout.h"
 #include "zen/scene/view.h"
 #include "zen/wlr/box.h"
-#include "zen/xdg-toplevel-view.h"
 
 struct surface_callback_data {
   zn_screen_for_each_visible_surface_callback_t callback;
--- a/zen/scene/screen.c
+++ b/zen/scene/screen.c
@@ -51,46 +51,25 @@ zn_screen_for_each_visible_surface(struct zn_screen *self,
 
 struct wlr_surface *
 zn_screen_get_surface_at(struct zn_screen *self, double x, double y,
-    double *surface_x, double *surface_y)
+    double *surface_x, double *surface_y, struct zn_view **view)
 {
-  struct zn_view *view;
+  struct zn_view *view_iterator;
   struct zn_board *board = zn_screen_get_current_board(self);
   double view_sx, view_sy;
   struct wlr_surface *surface;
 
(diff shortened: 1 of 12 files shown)


zen/scene/screen.c
```diff
@@ -6,23 +6,51 @@
 #include "zen/scene/screen-layout.h"
 #include "zen/scene/view.h"
 #include "zen/wlr/box.h"
+#include "zen/xdg-toplevel-view.h"
+
+static void
+send_frame_done(struct wlr_surface *surface, void *data)
+{
+  wlr_surface_send_frame_done(surface, data);
+}
```
watasuke102, line 15: Do you think this function will update later? If not, this function is now just alias, so I suggest it should remove.

shierote: It may be updated, but right now this function is redundant. I was following past code conventions, but I thought it should be removed too. I will fix!

zen/scene/screen.c
```diff
@@ -7,41 +7,88 @@
...
+    surface = view->impl->get_wlr_surface_at(
+        view, view_sx, view_sy, surface_x, surface_y);
+    if (surface != NULL) return surface;
   }
+
+  return NULL;
 }
 
 struct zn_view *
 zn_screen_get_view_at(
```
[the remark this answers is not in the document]
shierote: This looks good. I wrote zn_screen_get_surface_at in this pr, so I will modify it.

watasuke102 (approved):
LGTM!

Aki-7 (approved):
LGTM!

Issue #167, opened 2022-09-21
Segmentation Fault due to access invalid seat


watasuke102:
If exiting zen when some view is focused, caused Segmentation Fault like following:

```
0x000055555555f717 in zn_scene_set_focused_view (self=0x555555d83ea0, view=0x0) at ../zen/scene/scene.c:78
78        struct wlr_seat* seat = server->input_manager->seat->wlr_seat;
#0  0x000055555555f717 in zn_scene_set_focused_view (self=0x555555d83ea0, view=0x0) at ../zen/scene/scene.c:78
#1  0x00005555555609a8 in wl_signal_emit (data=0x0, signal=0x555555f4dde8) at /usr/include/wayland-server-core.h:484
#2  zn_view_unmap (self=self@entry=0x555555f4dda0) at ../zen/scene/view.c:165
#3  0x000055555555dadc in zn_xdg_toplevel_view_handle_unmap (listener=0x555555f4de30, data=<optimized out>) at ../zen/xdg-toplevel-view.c:81
#4  0x00007ffff7d6ec14 in wlr_signal_emit_safe (signal=0x5555555a4eb8, data=0x5555555a4db0) at ../subprojects/wlroots/util/signal.c:29
#5  0x00007ffff7d3cf01 in unmap_xdg_surface (surface=0x5555555a4db0) at ../subprojects/wlroots/types/xdg_shell/wlr_xdg_surface.c:40
#6  0x00007ffff7d3dcc4 in reset_xdg_surface (xdg_surface=0x5555555a4db0) at ../subprojects/wlroots/types/xdg_shell/wlr_xdg_surface.c:426
#7  0x00007ffff7d3de12 in destroy_xdg_surface (surface=0x5555555a4db0) at ../subprojects/wlroots/types/xdg_shell/wlr_xdg_surface.c:464
#8  0x00007ffff7d3d9e8 in xdg_surface_handle_surface_destroy (listener=0x5555555a4e48, data=0x555555f4ce80) at ../subprojects/wlroots/types/xdg_shell/wlr_xdg_surface.c:361
#9  0x00007ffff7d6ec14 in wlr_signal_emit_safe (signal=0x555555f4d150, data=0x555555f4ce80) at ../subprojects/wlroots/util/signal.c:29
#10 0x00007ffff7d6002f in surface_handle_resource_destroy (resource=0x55555610c7a0) at ../subprojects/wlroots/types/wlr_surface.c:702
[...]

Issue #153, opened 2022-09-11
Initialize `zn_scene::unmap_focused_view_listener` on create `zn_scene`


Aki-7:
## Reproduction

- Start `zen-desktop` with no argument
- Do not start any view
- Stop `zen-desktop` gracefully (<alt> + q)
- Segmentation fault

## Cause
- `zn_scene::unmap_focused_view_listener` is not initialized in `zn_scene_create`

watasuke102:
solved by #168

PR #168, opened 2022-09-21
[refactor] destroying process

watasuke102:
## Context

See #167. And now, Segmentation Fault occurs when destroying texture.

## Summary

destroy resources (e.g. client, backend, texture) before destroying server.

## How to check behavior

start zen and exit, check no error occured. Leave comments if you expect to destroy something before exiting.

### DIFF
--- a/include/zen/cursor.h
+++ b/include/zen/cursor.h
@@ -49,4 +49,6 @@ void zn_cursor_set_xcursor(struct zn_cursor* self, const char* name);
 
 struct zn_cursor* zn_cursor_create(void);
 
+void zn_cursor_destroy_resources(struct zn_cursor* self);
+
 void zn_cursor_destroy(struct zn_cursor* self);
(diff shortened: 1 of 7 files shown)


Aki-7:
When I terminate zen without starting any client, it causes a segmentation fault. Please let me know if you can't reproduce this.
```log
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7dcceaf in wl_list_remove () from /usr/lib/libwayland-server.so.0
#0  0x00007ffff7dcceaf in wl_list_remove () at /usr/lib/libwayland-server.so.0
#1  0x000055555555ec6c in zn_scene_destroy (self=0x5555556e12d0) at ../zen/scene/scene.c:278
#2  0x000055555555c6ec in zn_server_destroy (self=0x55555558a580) at ../zen/server.c:356
```

watasuke102:
That change will solve #153

shierote (approved):
In my env, it works well.

Aki-7 (approved):

PR #152, opened 2022-09-11
Add background images

liwii:
## Context
In the current `zen-desktop`, the background is filled with one green-ish color. With a background image, the system would be more attractive to the users.
## Summary
Install the background images provided by @rtr-dnd and draw it on the screen. If the screen size ratio does not match the image size ratio, the image is scaled with aspect-fill manner.
image_123927839
## How to check behavior
Run `zen-desktop -s weston-terminal` and see if the background image is properly set. You can also test the other bundled background images by changing the line in https://github.com/zigen-project/zen/compare/main...liwii:bg?expand=1#diff-d6c974d638e0e2ea1f88ab4bb8e67b78b22a6aa636536a72467d9b39bb9defe0R14.  
If needed, change the screen orientation and the scale like the comment in #121 and test it again.

Issue #159, opened 2022-09-19
Make other formats of images (jpg, rgb-png, ..etc) available as the background image.


liwii:
Now `zn_scene_setup_background` only accepts PNG images with RGBA format. It would be better if we could use other widely-used formats.

liwii:
SVG images have a [corresponding surface creation function](https://www.cairographics.org/manual/cairo-SVG-Surfaces.html#cairo-svg-surface-create), but we need to treat the size of surface differently. JPEG images do not even have such a function. I'll deal with these formats in other PRs later.

PR #169, opened 2022-09-25
Bg image fix

rtr-dnd:
## Context

The background image set by @liwii at #152 was different from the one we decided to use as the default. 
In addition, there were a few glitches in other bundled images too, such as transparency problem and shapes not aligning properly.

## Summary

I re-exported bg images to fix those problems. File names are either changed or swapped. Now main bg is the one with gradient and white mesh.


watasuke102 (changes requested):
Background image isn't shown, and following error log is shown:

```
[ERROR] [zen/scene/scene.c:202] Image format not supported
```

This problem will may solve (#159), but for now, please convert image file. ([ref](https://github.com/zigen-project/zen/pull/160#issuecomment-1250618974))

rtr-dnd:
@watasuke102 
Thank you for your review; I was not aware of the problem.
The images are now converted to RGBA files.

watasuke102 (approved):
Nice change!

Aki-7 (approved):
```
!!!            !!!!!!!!!!!!    !!!!!!!!!!!!!!!     !!!!!!   !!!!!!       !!! 
!!!            !!!                  !!!           !!! !!!  !!! !!!       !!!
!!!            !!!    !!!!!!        !!!           !!!  !!! !!!  !!!      !!!
!!!            !!!       !!!        !!!          !!!   !!!!!!   !!!
!!!!!!!!!!!!   !!!!!!!!!!!!         !!!          !!!    !!!!!    !!!     !!!
```

PR #171, opened 2022-09-27
Add popup of popup

shierote:
## Context

issue: #116 
before: #165

## Summary

https://github.com/zigen-project/zen/pull/171/commits/9db5739fdd685a551714925fc988f383a3b9f2ea
- add support for cases where a popup is generated under a popup (popup tree).

https://github.com/zigen-project/zen/pull/171/commits/f94dfb5642f427819e0e9ca11b69aa67db501a66
- remove unnecessary NULL check
  - In the `wlroots (0.15)` implementation, when the parent `xdg_surface` is deleted, the popups under it are also deleted.
  - Therefore, there is no need to consider the case where view_child does not have a parent view.

## How to check behavior
The following behaviors can be checked with the Wayland clients which has popup tree (such as `vlc`)

- popup of popup's damage works properly
  - turns blue when hovered over
  - returns to its original color when unhovered
  - hover/unhover a popup outside the frame of the view.
- popup of popup's input works properly
  - function works (ex. window appears) when hovered and clicked
  - If you unhover and click on that area, nothing comes back
  - click a popup outside the frame of the view.
- same as other requirements for popups
  - `How to check befavior` #157 
  - `How to check befavior` #165


watasuke102 (approved):
_LGTM_

zen/scene/view-child.c
```diff
@@ -128,20 +104,8 @@ zn_view_child_init(struct zn_view_child *self,
...
-      NULL;  // TODO: handle if view_child's parent(view_child) exists.
-
-  self->mapped = false;
 
-  self->view_unmap_listener.notify = zn_view_child_handle_view_unmap;
-  wl_signal_add(&self->view->events.unmap, &self->view_unmap_listener);
-  self->view_destroy_listener.notify = zn_view_child_handle_view_destroy;
-  wl_signal_add(&self->view->events.destroy, &self->view_destroy_listener);
-}
+  self->parent = NULL;  // TODO: handle subsurface
```
shierote, line 108: `parent` member is used for subsurface in the near future.

Aki-7 (approved):
```
  _    ___ _____ __  __ _ 
 | |  / __|_   _|  \/  | |
 | |_| (_ | | | | |\/| |_|
 |____\___| |_| |_|  |_(_)
```

PR #172, opened 2022-09-27
Render decoration

watasuke102:
## Context

On the x client of some of the wayland client, resizing / moving isn't available. This problem is solved to implement "decoration"; the border of the window.

## Summary

imprement rendering decoration (only drawing; cannot interaction)

## How to check behavior

start zen, and start some client. If the client doesn't have client-side decoration, the server-side decoration is rendered.

Screenshot 2022-09-28 00-01-08


shierote:
If you could indicate which parts of this picture are client-side decoration and which parts are server-side decoration, that would be great.

watasuke102:
server-side decoration is renderered with the color #0e1e4d (dark blue, used at zen default background). In that image, `weston-terminal` and `vlc` use client-side decoration, and `xcalc` use server-side decoration.

zen/decoration-manager.c
```diff
@@ -0,0 +1,66 @@
...
+err:
+  return NULL;
+}
+
+void
+zn_decoration_manager_destroy(struct zn_decoration_manager* self)
+{
+  struct zn_xdg_decoration *deco, *tmp;
+  wl_list_for_each_safe (deco, tmp, &self->xdg_decoration_list, link) {
+    zn_xdg_decoration_destory(deco);
```
Aki-7, line 62: [typo]
```suggestion
    zn_xdg_decoration_destroy(deco);
```

Aki-7: Do we need to destroy this here?
Destroying `zn_xdg_decoration` together with `wlr_xdg_toplevel_decoration_v1` seems enough.

Aki-7 (commented):
- `alacritty` doesn't work well. But it seems the problem of subsurface. You can leave this.

Aki-7 (commented):
Google Chrome has a server side decoration. Is this expected behavior?

watasuke102:
> alacritty doesn't work well.

What's happen?

> Google Chrome has a server side decoration

That is expected behavior. It's able to choose to use decoration from server-side or client-side. Try right click at the top of the window on other DE (because zen can't show x client's popup now).

watasuke102:
`alacritty` (and `kitty`) has client-side decoration consisting of subsurface, at the outside of the toplevel surface. Currently, its damage isn't handled, but if subsurface damage is implemented (ref: #116), it will work well.

zen/xdg-decoration.c
```diff
@@ -0,0 +1,89 @@
...
+#include "zen/xdg-toplevel-view.h"
+
+static void
+zn_xdg_decoration_set_view_decoration(struct zn_xdg_decoration* self)
+{
+  self->view->requested_client_decoration =
+      self->wlr_decoration->requested_mode !=
+      WLR_XDG_TOPLEVEL_DECORATION_V1_MODE_SERVER_SIDE;
+  wlr_xdg_toplevel_decoration_v1_set_mode(
+      self->wlr_decoration, WLR_XDG_TOPLEVEL_DECORATION_V1_MODE_CLIENT_SIDE);
```
Aki-7, line 14: What if the client request server side decoration?

zen/screen-renderer.c
```diff
@@ -177,6 +177,71 @@ zn_screen_renderer_render_view_popups(struct zn_screen *screen,
...
+  zn_view_get_decoration_fbox(view, &fbox);
+
+  zn_output_box_effective_to_transformed_coords(
+      output, &fbox, &transformed_box);
+
+  pixman_region32_init(&render_damage);
+  pixman_region32_union_rect(&render_damage, &render_damage, transformed_box.x,
+      transformed_box.y, transformed_box.width, transformed_box.height);
+
+  pixman_region32_intersect(&render_damage, &render_damage, screen_damage);
```
Aki-7, line 202: I think it's better to avoid having the same variable name among function scope and inner block scope.
(`fbox` and `transformed_box`)

How about adding an additional inner scope?
```suggestion
{
  struct wlr_fbox fbox;
  struct wlr_box transformed_box;
  zn_view_get_decoration_fbox(view, &fbox);

  zn_output_box_effective_to_transformed_coords(
      output, &fbox, &transformed_box);

  pixman_region32_init(&render_damage);
  pixman_region32_union_rect(&render_damage, &render_damage, transformed_box.x,
      transformed_box.y, transformed_box.width, transformed_box.height);

  pixman_region32_intersect(&render_damage, &render_damage, screen_damage);
}
```

watasuke102: > avoid having the same variable name

That's true, so I will use different name in that scope

zen/xdg-decoration.c
watasuke102: I thought the client which has the client-side decoration should always draw it. But on second thought, ignoring request is not so good. (I wanted to try this behavior with Google Chrome later, but I found good client that can change decoration mode)

PR #174, opened 2022-09-30
change meaning of view's position

watasuke102:
## Context

See #147

## Summary

Change meaning of view::x, y. It pointing to surface's pos before, but now `geometry` position.

## How to check behavior

Move the view to some position, check that there is no difference between specified pos and appeared pos.

[suggest] insert following line to `view.c:191` (at: `zn_view_map_to_scene`)

```c
zn_view_move(self, NULL, 0, 0);
```

### DIFF
--- a/include/zen/scene/view-child.h
+++ b/include/zen/scene/view-child.h
@@ -8,7 +8,8 @@ struct zn_view_child;
 
 struct zn_view_child_impl {
   struct wlr_surface *(*get_wlr_surface)(struct zn_view_child *child);
-  void (*get_view_coords)(struct zn_view_child *child, int *sx, int *sy);
+  void (*get_toplevel_coords)(struct zn_view_child *child, double popup_sx,
+      double popup_sy, double *toplevel_sx, double *toplevel_sy);
 };
 
 /*
--- a/include/zen/scene/view-child.h
+++ b/include/zen/scene/view-child.h
@@ -8,8 +8,8 @@ struct zn_view_child;
 
 struct zn_view_child_impl {
   struct wlr_surface *(*get_wlr_surface)(struct zn_view_child *child);
-  void (*get_toplevel_coords)(struct zn_view_child *child, double popup_sx,
-      double popup_sy, double *toplevel_sx, double *toplevel_sy);
+  void (*get_toplevel_coords)(struct zn_view_child *child, double child_sx,
+      double child_sy, double *toplevel_sx, double *toplevel_sy);
 };
 
 /*
--- a/zen/input/cursor-grab-resize.c
+++ b/zen/input/cursor-grab-resize.c
@@ -115,7 +115,7 @@ zn_cursor_grab_resize_create(
   }
 
   struct wlr_fbox box;
-  zn_view_get_window_fbox(view, &box);
+  zn_view_get_view_fbox(view, &box);
 
   self->init_view_x = view->x;
   self->init_view_y = view->y;
--- a/zen/input/cursor-grab-resize.c
+++ b/zen/input/cursor-grab-resize.c
@@ -117,8 +117,6 @@ zn_cursor_grab_resize_create(
   struct wlr_fbox box;
   zn_view_get_view_fbox(view, &box);
 
-  self->init_view_x = view->x;
-  self->init_view_y = view->y;
   self->init_view_width = box.width;
   self->init_view_height = box.height;
   self->init_cursor_x = cursor->x;
--- a/zen/scene/screen.c
+++ b/zen/scene/screen.c
@@ -59,10 +59,12 @@ zn_screen_get_surface_at(struct zn_screen *self, double x, double y,
   struct wlr_surface *surface;
 
   wl_list_for_each_reverse (view_iterator, &board->view_list, link) {
-    double sx, sy;
-    view_sx = x - view_iterator->x;
-    view_sy = y - view_iterator->y;
+    struct wlr_fbox fbox;
+    zn_view_get_surface_fbox(view_iterator, &fbox);
+    view_sx = x - fbox.x;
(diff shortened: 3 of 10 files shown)


Aki-7 (commented):
- `zn_view_get_decoration_fbox` now seems to be able to replace with `zn_view_get_view_fbox`.

- Don't we have to add any changes to `zn_view_child`? If so, why? `zn_view_child` uses `view->x` internally.

zen/scene/view-child.c
```diff
@@ -29,12 +29,15 @@ zn_view_child_get_surface_fbox(
     struct zn_view_child *self, struct wlr_fbox *fbox)
 {
   struct wlr_surface *surface = self->impl->get_wlr_surface(self);
-  int sx, sy;
 
+  struct wlr_fbox view_fbox;
+  zn_view_get_surface_fbox(self->view, &view_fbox);
+
+  int sx, sy;
   self->impl->get_view_coords(self, &sx, &sy);
```
Aki-7, line 37: This doesn't get "view coordinates" any more. I thinks we should change the name.

zen/scene/screen.c
```diff
@@ -59,10 +59,12 @@ zn_screen_get_surface_at(struct zn_screen *self, double x, double y,
   struct wlr_surface *surface;
 
   wl_list_for_each_reverse (view_iterator, &board->view_list, link) {
-    double sx, sy;
-    view_sx = x - view_iterator->x;
-    view_sy = y - view_iterator->y;
+    struct wlr_fbox fbox;
+    zn_view_get_surface_fbox(view_iterator, &fbox);
```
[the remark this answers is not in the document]
watasuke102: This is correct, because `zn_screen_get_surface_at` picks up the **surface** literally. `view_box` includes the outside of the surface (e.g. server-side decoration), it's not correct in this circumstance.

include/zen/scene/view-child.h
```diff
@@ -8,7 +8,12 @@ struct zn_view_child;
 
 struct zn_view_child_impl {
   struct wlr_surface *(*get_wlr_surface)(struct zn_view_child *child);
-  void (*get_view_coords)(struct zn_view_child *child, int *sx, int *sy);
+  /**
+   * get relative pos from:
+   *  view has client-side decoration  -> view pos
+   *  otherwise (use server-side deco) -> surface pos
+   */
+  void (*get_relative_pos)(struct zn_view_child *child, int *sx, int *sy);
```
watasuke102, line 16: This naming doesn't seem so good; does anyone have any good ideas?

zen/screen-renderer.c
```diff
@@ -100,8 +100,11 @@ zn_screen_renderer_get_surface_fbox(struct wlr_surface *surface,
     struct zn_view *view, int surface_x, int surface_y,
     struct wlr_fbox *surface_box)
 {
-  surface_box->x = view->x + surface_x;
-  surface_box->y = view->y + surface_y;
+  struct wlr_fbox fbox;
```
Aki-7, line 103: This function (`zn_screen_renderer_get_surface_fbox`) does not seem to be generic at all..
Could you remove this function and inline it in `zn_screen_renderer_for_each_surface_iterator`?

```c
static void
zn_screen_renderer_for_each_surface_iterator(
    struct wlr_surface *surface, int surface_x, int surface_y, void *user_data)
{
  struct surface_iterator_data *data = user_data;
  struct wlr_fbox surface_box;
  struct wlr_fbox fbox;

  zn_view_get_surface_fbox(data->view, &fbox);

  surface_box.x = fbox.x + surface_x;
  surface_box.y = fbox.y + surface_y;
  surface_box.width = surface->current.width;
  surface_box.width = surface->current.height;

  data->iterator(
      data->screen, surface, data->renderer, &surface_box, data->screen_damage);
}
```

zen/screen-renderer.c
```diff
@@ -203,12 +206,12 @@ zn_screen_renderer_render_decoration(struct zn_screen *screen,
 
   {
     pixman_region32_t region;
-    struct wlr_fbox window_fbox;
+    struct wlr_fbox surface_fbox;
     struct wlr_box transformed_window_box;
 
-    zn_view_get_window_fbox(view, &window_fbox);
+    zn_view_get_surface_fbox(view, &surface_fbox);
```
Aki-7, line 212: The name of "surface" and "window" is mixing (`surface_fbox` and `transformed_window_box`), and as you know, it's confusing part. How about using the name "window" and add comment about it?

Changing function name seems good, too.
`zn_screen_renderer_render_decoration` -> `zn_screen_renderer_render_server_decoration`

```suggestion
    zn_view_get_surface_fbox(view, &window_fbox); // Using server-side decoration, window_fbox and surface_fbox is the same
```

zen/input/cursor-grab-resize.c
```diff
@@ -115,7 +115,7 @@ zn_cursor_grab_resize_create(
   }
 
   struct wlr_fbox box;
-  zn_view_get_window_fbox(view, &box);
+  zn_view_get_view_fbox(view, &box);
 
   self->init_view_x = view->x;
   self->init_view_y = view->y;
```
Aki-7, line 121: Not related to this pull request, but these properties are not used. (`init_view_(x|y)`)
Could you remove them?

Aki-7 (approved):
Kudos to the great refactoring!

```
                                  ,---. 
,--.    ,----.,--------.,--.   ,--.|   | 
|  |   '  .-./'--.  .--'|   `.'   ||  .' 
|  |   |  | .---.|  |   |  |'.'|  ||  |  
|  '--.'  '--'  ||  |   |  |   |  |`--'  
`-----' `------' `--'   `--'   `--'.--.  
                                   '--'  
```

PR #177, opened 2022-10-05
Add shortcuts for view alignment

liwii:
## Context
Some applications like [Magnet on MacOS](https://magnet.crowdcafe.com/) allow users to align the views with keyboard. As a developer-oriented windowing system, it is better for zen-desktop to have such a functionality by default.

## Summary
Added keyboard shortcuts  to align the focused view & move focuses.

## How to check behavior
Build zen-desktop & try the shortcuts added to the README.


README.adoc
```diff
@@ -70,4 +70,25 @@ We cannot customize these bindings for now.
 |<alt> + q
 |Terminate ZEN Desktop (for development convenience)
 
+|<alt> + <ctrl> + f
+|Make the focused view full screen
```
watasuke102, line 74: You mean this is `maximize`? If so, I want you to follow [the protocol](https://wayland.app/protocols/xdg-shell#xdg_toplevel:request:set_maximized).

But this task doesn't seem easy, so if you won't do this, please remove this feature.

watasuke102 (commented):
Most of the shortcuts worked well. I haven't read the code yet, so I leave the comment about the behavior of the shortcut.

README.adoc
liwii: Thanks for the suggestion, I deleted the `maximize` key binding. I also fixed the other key bindings so that they work similarly as [Ubuntu does](https://help.ubuntu.com/stable/ubuntu-help/shell-keyboard-shortcuts.html.en).

watasuke102 (approved):
LGTM!

zen/scene/scene.c
```diff
@@ -51,6 +51,66 @@ zn_scene_move_board_binding_handler(
...
+
+static void
+zn_scene_align_focused_view_binding_handler(
+    uint32_t time_msec, uint32_t key, void* data)
+{
+  struct zn_scene* scene = data;
+  UNUSED(time_msec);
+
+  if (scene->focused_view == NULL) return;
+  struct zn_board* board = scene->focused_view->board;
```
Aki-7, line 95: We have a suitable function.
The same is applied to `zn_scene_switch_focused_view_binding_handler`
```suggestion
  struct zn_board* board = zn_scene_get_board(scene);
  if (board == NULL) { ... }
```

zen/scene/scene.c
```diff
@@ -51,6 +51,66 @@ zn_scene_move_board_binding_handler(
...
+
+  if (scene->focused_view == NULL) return;
+  // Should be mapped
+  struct zn_board* focused_board = scene->focused_view->board;
+  if (focused_board == NULL) {
+    zn_error("Something went wrong");
+    return;
+  }
+  if (wl_list_length(&focused_board->view_list) == 1) {
+    zn_warn("Only 1 view");
```
Aki-7, line 69: I thinks it's a normal situation and doesn't need warning.
And perhaps we should focus to the only one view since the view can be not focused yet.
(It's possible that there is no focused view)

zen/scene/scene.c
```diff
@@ -51,6 +51,66 @@ zn_scene_move_board_binding_handler(
...
+    return;
+  }
+
+  struct zn_view* new_focused_view;
+  if (key != KEY_TAB) {
+    zn_error("Invalid key: %u", key);
+    return;
+  }
```
Aki-7, line 77: I prefer to use `zn_assert`.
I think assertion is a good way to indicate a programmatically unreachable situation.

```suggestion
  if (!zn_assert(key == KEY_TAB, "Invalid key: %u", key)) {
    return;
  }
```

The same can be applied to `zn_scene_align_focused_view_binding_handler`

zen/scene/scene.c
```diff
@@ -51,6 +51,66 @@ zn_scene_move_board_binding_handler(
   }
 }
 
+static void
+zn_scene_switch_focused_view_binding_handler(
```
Aki-7, line 55: We decided to use `handle_*` instead of `*_handler`.
https://github.com/zigen-project/zen/pull/157#issuecomment-1250365166
https://github.com/zigen-project/zen/commit/b9d2cc51553efc4466d9ccfd0fdd81ddd982f838

Could you use `handle_*` and fix `*_handler` functions in this file?

```suggestion
zn_scene_handle_switch_focused_view_binding(
```

zen/scene/scene.c
```diff
@@ -51,6 +51,66 @@ zn_scene_move_board_binding_handler(
...
+    zn_warn("Only 1 view");
+    return;
+  }
+
+  struct zn_view* new_focused_view;
+  if (key != KEY_TAB) {
+    zn_error("Invalid key: %u", key);
+    return;
+  }
+  // bring prev to front (should be the second last)
```
Aki-7, line 78: > (should be the second last)

What do you mean by this?

liwii: I tried to say that the view in the second position counted from the last is brought to the last in the list.
I now think that I do not need to explain it, so I simply removed the comment.

liwii:
Sorry for leaving the PR for a while, but now I think all the points are fixed.

(and sorry for the messy commit log due to something about `rebase` ...)

PR #178, opened 2022-10-05
introduce remote display system

Aki-7:
## Context

Introduce remote immersive display system

## Summary

- [x] implement display system switch
  - switch display system using `cui-client`
  - stop updating screen display system when using immersive display system
  - stop cursor event when using immersive display system
- [x] introduce building `remote-scene`
  - `remote-scene` consists of `remote-object`s
  - `zn_board_remote_object` is the first `remote-object`
  - `remote-scene` is basically mirror of `zn_scene`. `zn_scene` represents pure logic, and `remote-scene` constructs the information necessary for drawing, such as meshes, from the properties of `zn_scene`.

Any known unfinished work is left as TODO comments.

## How to check behavior

1. Build and run
2. Start `cui-client` and type "immersive". After that, the screen will not be updated and cursor will be unavailable.
3. During "immersive" display mode, the count of `zn_board` and `zn_board_remote_object` will be synchronized.

You can check the behavior by adding debugging code shown below.

```c
static void
zn_remote_scene_handle_new_board(struct wl_listener* listener, void* data)
{
  struct zn_remote_scene* self =
      zn_container_of(listener, self, new_board_listener);

  struct zn_board* board = data;

  zn_board_remote_object_create(board, self);

  zn_info("%d boards", wl_list_length(&self->board_object_list)); // HERE
}
```

```c
static void
zn_remote_scene_clean_build(struct zn_remote_scene* self)
{
  struct zn_board* board;

  zn_remote_scene_clean(self);

  wl_list_for_each (board, &self->scene->board_list, link) {
    zn_board_remote_object_create(board, self);
  }
  zn_info("%d boards", wl_list_length(&self->board_object_list)); // HERE
}
```

### DIFF
--- a/include/zen/immersive/remote/object/board.h
+++ b/include/zen/immersive/remote/object/board.h
@@ -1,0 +1,19 @@
+#pragma once
+
+#include <wayland-server-core.h>
+
+#include "zen/immersive/remote/scene.h"
+#include "zen/scene/board.h"
+
+struct zn_board_remote_object {
+  struct zn_board* board;  // non null;
+
+  struct wl_list link;  // zn_remote_immersive_render::scene::board_object_list
+
+  struct wl_listener board_destroy_listener;
+};
+
+struct zn_board_remote_object* zn_board_remote_object_create(
+    struct zn_board* board, struct zn_remote_scene* remote_scene);
+
+void zn_board_remote_object_destroy(struct zn_board_remote_object* self);
--- a/include/zen/immersive/display-system.h
+++ b/include/zen/immersive/display-system.h
@@ -1,0 +1,18 @@
+#pragma once
+
+#include <wayland-server-core.h>
+
+#include "zen/immersive/renderer.h"
+
+struct zn_immersive_display_system {
+  struct {
+    struct wl_signal activate;
+    struct wl_signal deactivated;
+  } events;
+};
+
+struct zn_immersive_display_system* zn_immersive_display_system_create(
+    struct wl_display* display, struct zn_immersive_renderer* renderer);
+
+void zn_immersive_display_system_destroy(
+    struct zn_immersive_display_system* self);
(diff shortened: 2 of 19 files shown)


include/zen/immersive/remote/object/board.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include <wayland-server-core.h>
+
+#include "zen/immersive/remote/scene.h"
+#include "zen/scene/board.h"
+
+struct zn_board_remote_object {
```
watasuke102, line 8: This struct seems to belong to the namespace `zen::remote_object::board`, so the name should be `zn_remote_object_board`. This name is more convenient for completion, and easy to understand at a glance. Also file name should be changed.

Aki-7: > This struct seems to belong to the namespace `zen::remote_object::board`

yes, (precisely, `zen::remote::object::board` I think) but

> so the name should be `zn_remote_object_board`

I thinks this is a matter of discussion.

For example,
https://github.com/swaywm/wlroots/blob/master/include/render/gles2.h#L34
The renderer for GLES2 in wlroots can be considered to belong to the namespace `wlr::renderer::gles2` but has the name `wlr_gles2_renderer`.
Ultimately, it is a renderer with the gles2 attribute, so "renderer" should come at the end.

Similarly, `zn_board_remote_object` is ultimately a `remote_object`, so I think "remote_object" should come last.

> This name is more convenient for completion

Interesting point of view, but it's possible to have lot more `zn_remote_object*` in the future, so it does not seem critical reason to choose `zn_remote_object_board`.

> Also file name should be changed.

How? Some thing like `remote-object-board.h`?

watasuke102: Ok, I see. `wlr_glew2_renderer` is good example!

watasuke102 (approved):
Nice work!

Aki-7:
Thanks for review!

PR #181, opened 2022-10-08
Implement move/resize to x client by server-side decoration

watasuke102:
## Context

At #172 , server-side decoration is rendered, but it isn't able to interact (move/resize).

## Summary

Add `zn_screen_get_view_area_type_at`, and implement interaction to server-side decoration.

## How to check behavior

1. start zen with some x client
2. drag border of the server-side decoration (colored dark blue; #0e1e4d)

You can interact like other software or DE.

### DIFF
--- a/include/zen/cursor.h
+++ b/include/zen/cursor.h
@@ -38,6 +38,8 @@ void zn_cursor_start_grab(struct zn_cursor* self, struct zn_cursor_grab* grab);
 
 void zn_cursor_end_grab(struct zn_cursor* self);
 
+const char* zn_cursor_get_resize_xcursor_name(uint32_t edges);
+
 void zn_cursor_move_relative(struct zn_cursor* self, double dx, double dy);
 
 void zn_cursor_get_fbox(struct zn_cursor* self, struct wlr_fbox* fbox);
--- a/include/zen/cursor.h
+++ b/include/zen/cursor.h
@@ -38,7 +38,7 @@ void zn_cursor_start_grab(struct zn_cursor* self, struct zn_cursor_grab* grab);
 
 void zn_cursor_end_grab(struct zn_cursor* self);
 
-const char* zn_cursor_get_resize_xcursor_name(uint32_t edges);
+void zn_cursor_set_resizing_cursor(struct zn_cursor* self, uint32_t edges);
 
 void zn_cursor_move_relative(struct zn_cursor* self, double dx, double dy);
 
--- a/include/zen/scene/view.h
+++ b/include/zen/scene/view.h
@@ -64,7 +64,7 @@ struct zn_view {
   } resize_status;
 
   struct {
-    struct wl_signal surface_resized;
+    struct wl_signal surface_resized;  // (struct wlr_surface)
     struct wl_signal unmap;
     struct wl_signal destroy;
   } events;
--- a/include/zen/scene/view.h
+++ b/include/zen/scene/view.h
@@ -99,6 +99,12 @@ void zn_view_map_to_scene(struct zn_view *self, struct zn_scene *scene);
 
 void zn_view_unmap(struct zn_view *self);
 
+/**
+ * @retval bits sum of enum wlr_edges
+ * @param type bits sum of enum zn_view_area_type
+ */
+uint32_t zn_view_convert_area_type_to_wlr_edges(uint32_t type);
+
 void zn_view_init(struct zn_view *self, enum zn_view_type type,
     const struct zn_view_impl *impl);
 
--- a/include/zen/scene/view.h
+++ b/include/zen/scene/view.h
@@ -70,14 +70,6 @@ struct zn_view {
   } events;
 };
 
-/**
- * Add the damage of all surfaces associated with the view to the output where
- * the view is displayed.
- */
-void zn_view_damage(struct zn_view *self);
-
-void zn_view_damage_whole(struct zn_view *self);
-
 void zn_view_bring_to_front(struct zn_view *self);
 
 /**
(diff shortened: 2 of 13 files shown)


include/zen/scene/view.h
```diff
@@ -7,7 +7,7 @@
 
 #include "zen/scene/scene.h"
 
-#define VIEW_DECORATION_BORDER 3
+#define VIEW_DECORATION_BORDER 8
```
watasuke102, line 10: The size of border is updated, but if you feel it is difficult to start resizing, please leave comment.

zen/scene/view.c
```diff
@@ -108,6 +69,77 @@ zn_view_damage_whole(struct zn_view *self)
...
+{
+  wl_list_remove(&self->link);
+  wl_list_insert(self->board->view_list.prev, &self->link);
+
+  if (self->impl->restack) {
+    self->impl->restack(self, XCB_STACK_MODE_ABOVE);
+  }
+
+  zn_view_damage_whole(self);
+}
```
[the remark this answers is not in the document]
watasuke102: - `zn_view_update_pos_on_resizing` is similar to `zn_view_move` because it updates the position, so I want to place near `zn_view_move`.
- `zn_view_update_pos_on_resizing` is using `zn_view_add_damage_fbox`, but I don't like so much to add prototype declaration of the function

Therefore I moved them, but now `zn_view_update_pos_on_resizing()` is renamed to `zn_view_handle_surface_resized()` (It does not only update position!), and made static, so I'll revert this.

zen/xdg-toplevel-view.c
```diff
@@ -23,20 +23,7 @@ zn_xdg_toplevel_view_handle_wlr_surface_commit(
...
-  int dx = 0, dy = 0;
-  if (self->base.resize_status.edges & WLR_EDGE_LEFT) {
-    dx = surface->previous.width - surface->current.width;
-  }
-  if (self->base.resize_status.edges & WLR_EDGE_TOP) {
-    dy = surface->previous.height - surface->current.height;
-  }
-  zn_view_move(
-      &self->base, self->base.board, self->base.x + dx, self->base.y + dy);
+  wl_signal_emit(&self->base.events.surface_resized, data);
```
Aki-7, line 26: I think this signal doesn't need to have an argument.

```suggestion
  wl_signal_emit(&self->base.events.surface_resized, NULL);
```

include/zen/cursor.h
```diff
@@ -38,6 +38,8 @@ void zn_cursor_start_grab(struct zn_cursor* self, struct zn_cursor_grab* grab);
 
 void zn_cursor_end_grab(struct zn_cursor* self);
 
+const char* zn_cursor_get_resize_xcursor_name(uint32_t edges);
```
Aki-7, line 41: I want to avoid having method without its instance at the first param.

How about something like this?

```c
void zn_cursor_set_resizing_cursor(struct zn_cursor *cursor, uint32_t edges) {
  static const char* xcursor_names[] = {...};
  const char *xcursor_name = xcursor_names[edges];
  zn_cursor_set_xcursor(cursor, xcursor_name);
}
```

zen/cursor.c
```diff
@@ -20,24 +22,36 @@ default_grab_motion(
...
+
+    if (surface) {
+      wlr_seat_pointer_enter(seat, surface, surface_x, surface_y);
+      wlr_seat_pointer_send_motion(
+          seat, event->time_msec, surface_x, surface_y);
+      return;
+    }
+  }
+
+  if (type < ZN_VIEW_AREA_TYPE_BORDER_TOP) {
```
Aki-7, line 47: Using `weston-simple-egl`, when the cursor is near a left egde, the cursor image changes to the "left_side".
But once clicked, the view starts moving, not resizing. It happens only when the cursor is right near the boundary of the surface and border.

Aki-7 (commented):
I'm going out for a bit, so I'll submit my review, which is still in process.

zen/xwayland-view.c
```diff
@@ -12,13 +12,22 @@ static void
...
 
-  // FIXME: when resizing, adjust the position of the view
+  if (!self->base.resize_status.resizing) {
+    return;
+  }
+
+  wl_signal_emit(&self->base.events.surface_resized, NULL);
+
+  ++self->current_resize_serial;
+  if (self->current_resize_serial == self->base.resize_status.last_serial) {
```
Aki-7, line 28: Is it guaranteed that the number of times set_size is called and commit is called always match?
It seems difficult to guarantee it for me.

zen/input/cursor-grab-resize.c
```diff
@@ -122,6 +129,12 @@ zn_cursor_grab_resize_create(
   self->init_cursor_x = cursor->x;
   self->init_cursor_y = cursor->y;
 
+  if (!zn_view_has_client_decoration(view)) {
+    self->init_view_width -= VIEW_DECORATION_BORDER * 2;
+    self->init_view_height -=
+        VIEW_DECORATION_BORDER * 2 + VIEW_DECORATION_TITLEBAR;
```
Aki-7, line 135: it's not `view_(width|height)` any more. It seems strange for me.

zen/cursor.c
```diff
@@ -57,9 +75,17 @@ default_grab_button(
...
+        grab->cursor->screen, grab->cursor->x, grab->cursor->y, &view);
     zn_scene_set_focused_view(server->scene, view);
+
+    if (type == ZN_VIEW_AREA_TYPE_TITLEBAR) {
+      zn_cursor_grab_move_start(cursor, view);
+    }
+    if (type >= ZN_VIEW_AREA_TYPE_BORDER_TOP) {
+      const uint32_t edges = zn_view_convert_area_type_to_wlr_edges(type);
+      zn_cursor_grab_resize_start(cursor, view, edges);
+    }
```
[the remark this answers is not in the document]
watasuke102: > button events should not send to client

I think so, so I'll apply this change (for readability and leave intention). But, I don't think it is possible to interact to the server-side decoration while the seat pointer is focused to the surface.

PR #206, opened 2022-11-11
add znr_gl_base_technique

watasuke102:
## Context

add gl base technique to zen renderer

## Summary

added: `znr_gl_base technique` to `znr-remote`, `DrawArrays` to client's `GlBaseTechnique`.

(But now it can't be used)

## How to check behavior

none?


Aki-7 (approved):
LGTM!

PR #214, opened 2022-11-14
add znr_gl_base_technique_draw_arrays

watasuke102:
## Context

continued from #206 

## Summary

add znr_gl_base_technique_draw_arrays, send information about draw method

## How to check behavior

none

**This PR is based on #213**

### DIFF
--- a/zgnroot/include/zgnr/gl-base-technique.h
+++ b/zgnroot/include/zgnr/gl-base-technique.h
@@ -9,6 +9,7 @@ extern "C" {
 #endif
 
 enum zgnr_gl_base_technique_draw_method {
+  ZGNR_GL_BASE_TECHNIQUE_DRAW_NONE,
   ZGNR_GL_BASE_TECHNIQUE_DRAW_ARRAYS,
 };
 
@@ -35,6 +36,7 @@ struct zgnr_gl_base_technique {
 
     enum zgnr_gl_base_technique_draw_method draw_method;
     union zgnr_gl_base_technique_draw_args args;
+    bool draw_info_changed;
   } current;
 
   void *user_data;
--- a/zgnroot/include/zgnr/gl-base-technique.h
+++ b/zgnroot/include/zgnr/gl-base-technique.h
@@ -36,7 +36,7 @@ struct zgnr_gl_base_technique {
 
     enum zgnr_gl_base_technique_draw_method draw_method;
     union zgnr_gl_base_technique_draw_args args;
-    bool draw_info_changed;
+    bool draw_method_changed;
   } current;
 
   void *user_data;
--- a/zgnroot/include/zgnr/gl-base-technique.h
+++ b/zgnroot/include/zgnr/gl-base-technique.h
@@ -9,8 +9,8 @@ extern "C" {
 #endif
 
 enum zgnr_gl_base_technique_draw_method {
-  ZGNR_GL_BASE_TECHNIQUE_DRAW_NONE,
-  ZGNR_GL_BASE_TECHNIQUE_DRAW_ARRAYS,
+  ZGNR_GL_BASE_TECHNIQUE_DRAW_METHOD_NONE,
+  ZGNR_GL_BASE_TECHNIQUE_DRAW_METHOD_ARRAYS,
 };
 
 union zgnr_gl_base_technique_draw_args {
--- a/zgnroot/include/zgnr/gl-base-technique.h
+++ b/zgnroot/include/zgnr/gl-base-technique.h
@@ -8,6 +8,18 @@
 extern "C" {
 #endif
 
+enum zgnr_gl_base_technique_draw_method {
+  ZGNR_GL_BASE_TECHNIQUE_DRAW_ARRAYS,
+};
+
+union zgnr_gl_base_technique_draw_args {
+  struct {
+    uint32_t mode;
+    int32_t count;
+    int32_t first;
+  } arrays;
+};
+
 struct zgnr_gl_base_technique {
   struct zgnr_rendering_unit *unit;  // nonnull
 
@@ -20,6 +32,9 @@ struct zgnr_gl_base_technique {
   struct {
     struct zgnr_gl_vertex_array *vertex_array;  // nullable
     bool vertex_array_changed;
+
+    enum zgnr_gl_base_technique_draw_method draw_method;
+    union zgnr_gl_base_technique_draw_args args;
   } current;
 
   void *user_data;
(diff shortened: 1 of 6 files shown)


zgnroot/include/zgnr/gl-base-technique.h
```diff
@@ -8,6 +8,18 @@
 extern "C" {
 #endif
 
+enum zgnr_gl_base_technique_draw_method {
+  ZGNR_GL_BASE_TECHNIQUE_DRAW_ARRAYS,
+};
```
Aki-7, line 13:
```suggestion
enum zgnr_gl_base_technique_draw_method {
  ZGNR_GL_BASE_TECHNIQUE_DRAW_METHOD_NONE,
  ZGNR_GL_BASE_TECHNIQUE_DRAW_METHOD_ARRAYS,
};
```

We need an initial state of `zgnr_gl_base_technique::current.draw_method`, which means no drawing method specified yet.

zgnroot/src/gl-base-technique.c
```diff
@@ -118,10 +119,12 @@ zgnr_gl_base_technique_protocol_draw_arrays(struct wl_client *client,
     struct wl_resource *resource, uint32_t mode, int32_t first, uint32_t count)
 {
   UNUSED(client);
-  UNUSED(resource);
-  UNUSED(mode);
-  UNUSED(first);
-  UNUSED(count);
+  struct zgnr_gl_base_technique_impl *self =
+      wl_resource_get_user_data(resource);
+  self->pending.draw_method = ZGNR_GL_BASE_TECHNIQUE_DRAW_ARRAYS;
```
Aki-7, line 124:
```suggestion
  struct zgnr_gl_base_technique_impl *self =
      wl_resource_get_user_data(resource);
  if (self == NULL) return;
      
  self->pending.draw_method = ZGNR_GL_BASE_TECHNIQUE_DRAW_ARRAYS;
```

Regarding base technique, it's wl_resource can be inert.
Inert resource means, the resoruce object itself is alive but there is no bound object. That is, wl_resource_get_user_data(resource) is NULL.

As for base technique, that happens when rendering unit bound to the base technique is destroyed.

If you need more explanation, feel free to ask me at any time!

zgnroot/src/gl-base-technique.c
```diff
@@ -178,12 +181,23 @@ zgnr_gl_base_technique_handle_rendering_unit_commit(
...
-  if (vertex_array) {
-    zgnr_gl_vertex_array_commit(vertex_array);
-    zgnr_gl_base_technique_set_current_vertex_array(self, &vertex_array->base);
-  } else {
-    zgnr_gl_base_technique_set_current_vertex_array(self, NULL);
+  self->base.current.args = self->pending.args;
+  self->base.current.draw_method = self->pending.draw_method;
+
+  self->base.current.args = self->pending.args;
+  self->base.current.draw_method = self->pending.draw_method;
```
Aki-7, line 188: DRY :)

zna/scene/virtual-object/gl-base-technique.c
```diff
@@ -58,9 +58,22 @@ zna_gl_base_technique_apply_commit(
...
 
-  // TODO: Apply draw arrays, etc.
+  switch (self->zgnr_gl_base_technique->current.draw_method) {
+    case ZGNR_GL_BASE_TECHNIQUE_DRAW_ARRAYS:
+      znr_gl_base_technique_draw_arrays(self->znr_gl_base_technique,
+          self->zgnr_gl_base_technique->current.args.arrays.mode,
+          self->zgnr_gl_base_technique->current.args.arrays.first,
+          self->zgnr_gl_base_technique->current.args.arrays.count);
+      break;
+  }
```
Aki-7, line 76: Summing up above discussion, we need to

1. check that the current draw method has been changed or not.
2. be careful of the initial state

You also have to consider about the situation where only_damaged == false, which means session has been reconnected and we need to do `znr_gl_base_technique_draw_ararys` or similar even if the draw method has not been modified.

zgnroot/include/zgnr/gl-base-technique.h
```diff
@@ -8,6 +8,19 @@
 extern "C" {
 #endif
 
+enum zgnr_gl_base_technique_draw_method {
+  ZGNR_GL_BASE_TECHNIQUE_DRAW_NONE,
+  ZGNR_GL_BASE_TECHNIQUE_DRAW_ARRAYS,
```
Aki-7, line 13:
```suggestion
  ZGNR_GL_BASE_TECHNIQUE_DRAW_METHOD_NONE,
  ZGNR_GL_BASE_TECHNIQUE_DRAW_METHOD_ARRAYS,
```

zgnroot/include/zgnr/gl-base-technique.h
```diff
@@ -20,6 +33,10 @@ struct zgnr_gl_base_technique {
   struct {
     struct zgnr_gl_vertex_array *vertex_array;  // nullable
     bool vertex_array_changed;
+
+    enum zgnr_gl_base_technique_draw_method draw_method;
+    union zgnr_gl_base_technique_draw_args args;
+    bool draw_info_changed;
```
Aki-7, line 39:
```suggestion
    bool draw_method_changed;
```

"draw info" seems a bit too broad in meaning. This could include vertex array, texture, etc.

zgnroot/src/gl-base-technique.c
```diff
@@ -238,11 +250,14 @@ zgnr_gl_base_technique_create(struct wl_client *client, uint32_t id,
...
   self->base.comitted = false;
+  self->base.current.draw_method = ZGNR_GL_BASE_TECHNIQUE_DRAW_NONE;
+  self->base.current.draw_info_changed = false;
+  self->pending.draw_info_changed = false;
+
   self->base.current.vertex_array = NULL;
   self->base.current.vertex_array_changed = false;
 
   wl_signal_init(&self->base.events.destroy);
-
```
Aki-7, line 245: Why did you remove this line? It's just a mistake or any intension?

zgnroot/include/zgnr/gl-base-technique.h
watasuke102: > "draw info" seems a bit too broad in meaning

Yes, I agree this. but I think when call `draw_array` twice, but draw method isn't changed.
So I named this `draw_info` instead of `draw_method` to inform that this monitors both method and args.

zgnroot/src/gl-base-technique.c
watasuke102: This is my mistake; occured when rebasing to main, I guess

Aki-7 (approved):

PR #259, opened 2022-12-12
fix hang on termination

Aki-7:
## Context

zen-desktop hangs on termination.

## Summary

Fix the issue.

## How to check behavior

zen-desktop will not hang on termination.


Aki-7:
@shierote Thanks!

PR #260, opened 2022-12-12
use prebuild grpc-dev to shorten the build time

Aki-7:
## Context

It taks too much time to build zen and other components which use grpc-dev.

## Summary

- [x] Use prebuild grpc-dev to shorten the build time.

## How to check behavior

Build time should be improved.


shierote (approved):

shierote:
LGTM

PR #270, opened 2022-12-16
Pointer events in IDS

Aki-7:
## Context

Pointer events in immersive display system.

## Summary

Send enter / motion / leave / button events to wayland client.

## How to check behavior

start `weston-clickdot` in immersive display system


watasuke102 (approved):
**_LGTM_**

PR #273, opened 2022-12-19
down grab

Aki-7:
## Context

While some button is pressing, pointer focus should not move.

## Summary

Add down grab.

## How to check behavior

Press button in weston-terminal, move cursor outside of the view without releasing. Pointer events still sent to the weston-terminal


watasuke102 (approved):
Nice change!

PR #279, opened 2022-12-22
Send keyboard events

watasuke102:
## Context

re-implement #161

## Summary

focus seat_keyboard to the view's surface

## How to check behavior

1. start zen with any client (reccomend: `weston-terminal`)
2. press any key (try also: Tab, arrow key, C-u, etc)
3. keyboard events will be sent


Aki-7 (approved):
Look Great to me!

PR #280, opened 2022-12-22
add `zn_data_device_manager`

watasuke102:
## Context

Some client such as Google Chrome cannot be launched due to error message: `Gdk: gdk_seat_get_keyboard: assertion 'GDK_IS_SEAT (seat)' failed`. It occurred because zen hasn't data device manager.

## Summary

Create `wlr_data_device_manager`. I also added `wlr_primary_selection_v1_device_manager`; make possible to paste selected text by wheel click.

## How to check behavior

1. start zen
2. try launch Google Chrome with these args: `--enable-features=UseOzonePlatform --ozone-platform=wayland` : Chrome starts up
4. Open some page with Chrome, select some text, and click wheel at text box : Selected text will be pasted

### DIFF
--- a/zen/data_device_manager.c
+++ b/zen/data_device_manager.c
@@ -1,0 +1,37 @@
+#include "zen/data_device_manager.h"
+
+#include <wlr/types/wlr_primary_selection_v1.h>
+#include <zen-common.h>
+
+struct zn_data_device_manager *
+zn_data_device_manager_create(struct wl_display *display)
+{
+  struct zn_data_device_manager *self;
+  self = zalloc(sizeof *self);
+  if (self == NULL) {
+    zn_error("Failed to allocate memory");
+    goto err;
+  }
+
+  self->wlr_data_device_manager = wlr_data_device_manager_create(display);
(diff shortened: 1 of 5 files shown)


zen/data_device_manager.c
```diff
@@ -0,0 +1,37 @@
...
+    goto err;
+  }
+
+  self->wlr_data_device_manager = wlr_data_device_manager_create(display);
+  if (self->wlr_data_device_manager == NULL) {
+    zn_error("Failed to create wlr_data_device_manager");
+    goto err_free;
+  }
+
+  wlr_primary_selection_v1_device_manager_create(display);
```
Aki-7, line 22: I know you don't need to have this return value as a member of zn_data_device_manager, but I want you to check that the value is not null.

zen/meson.build
```diff
@@ -1,6 +1,7 @@
 _zen_srcs = [
   'board.c',
   'cursor.c',
+  'data_device_manager.c',
```
Aki-7, line 4: we are using kebab case for file names.

Aki-7 (approved):

PR #284, opened 2022-12-26
Implement power button with zigzag

liwii:
## Summary
This PR adds UI widgets and `zigzag`, an internal library to implement UI widgets, to `zen`.

### zigzag
Currently, `zigzag` mainly has the following 4 features.
1. structures for UI widgets. Users can draw a tree under zigzag_layout with zigzag_nodes, and the top-level structure in the windowing system only needs to hold one zigzag_layout pointer.
1. pre- or post- process of texture drawing. Users only need to do drawing on cairo surface by registering render function.
1. Drawing utils for cairo surface. You can do several sets of actions such as drawing rectangles just by calling one util function.
1. Damage management. Instead of listening to all the UI nodes, users can register on_damage to zigzag_layout and get the notification from zigzag_layout if any of zigzag_node under that gets damaged.

### `zen/screen/zigzag-layout.c`
The file gives concrete implementation of the UI widgets. Currently,
- Half-transparent menu bar
- Non-clickable power button with the time displayed.
## How to check behavior
Confirm the menu-bar and the power button is shown in the display, and the time shown in the power button is updated minutely.


zen/screen/renderer.c
```diff
@@ -57,6 +58,8 @@ get_wallpaper_projection_matrix(struct wlr_output *output,
     box.height = wallpaper->height * output_width / wallpaper->width;
     box.y = (output_height - box.height) / 2;
   }
+  zn_error("Box: x = %d, y = %d, width = %d, height = %d", box.x, box.y,
+      box.width, box.height);
```
watasuke102, line 62: debug message?

include/zen/screen.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <wlr/util/box.h>
+#include <zigzag.h>
```
watasuke102, line 4: Why did you include this header at here?

zen/screen/renderer.c
```diff
@@ -155,6 +158,30 @@ render_view(struct zn_output *output, struct zn_view *view,
...
+    wlr_matrix_project_box(matrix, node->frame, WL_OUTPUT_TRANSFORM_NORMAL, 0,
+        output->wlr_output->transform_matrix);
+    int rect_count;
+    pixman_box32_t *rects =
+        pixman_region32_rectangles(screen_damage, &rect_count);
+
+    for (int i = 0; i < rect_count; i++) {
+      zn_error("Render a rect");
+      zn_error("x1: %d, y1: %d, x2: %d, y2: %d", rects[i].x1, rects[i].y1,
+          rects[i].x2, rects[i].y2);
```
watasuke102, line 177: debug message

watasuke102 (changes requested):
Please note that some debug message remains. I pointed out a part of them.

include/zen/screen.h
liwii: It's been left here since the prototype version is made. Now no need, thanks for pointing out!!

liwii:
Thanks for pointing them out. I think in the latest commit all the unnecessary debug messages are removed.

zen/screen/zigzag-layout.c
```diff
@@ -0,0 +1,280 @@
...
+
+static int
+zn_output_handle_minute_timer(void *data)
+{
+  struct zigzag_layout *self = (struct zigzag_layout *)data;
+  struct zn_zigzag_layout_state *state =
+      (struct zn_zigzag_layout_state *)self->state;
+  struct zn_server *server = zn_server_get_singleton();
+  long time_ms = current_time_ms();
+  state->next_min_ms += MSEC_PER_MIN;
```
Aki-7, line 29: As timer update is not always triggered every minutes in realtime clock (ex. in case the process stops with SIGSTOP), just adding 60 seconds is not safe. We should recalculate the next timing of `'00` sec again.

zen/screen/zigzag-layout.c
```diff
@@ -0,0 +1,280 @@
...
+};
+
+static int
+zn_output_handle_minute_timer(void *data)
+{
+  struct zigzag_layout *self = (struct zigzag_layout *)data;
+  struct zn_zigzag_layout_state *state =
+      (struct zn_zigzag_layout_state *)self->state;
+  struct zn_server *server = zn_server_get_singleton();
+  long time_ms = current_time_ms();
```
Aki-7, line 28: `current_time_ms` uses REALTIME CLOCK with which the clock can change significantly when, for example, NTP synchronization. We should use `clock_gettime(CLOCK_MONOTONIC, &ts);` in this case.

```suggestion
  struct timespec now;
  clock_gettime(CLOCK_MONOTONIC, &now);
  int64_t now_ms = timespec_to_msec(&now);
```

The same applies to the other use of `current_time_ms`.

zen/screen/zigzag-layout.c
```diff
@@ -0,0 +1,280 @@
...
+    .on_damage = zn_zigzag_layout_on_damage,
+};
+
+void
+power_button_on_click(struct zigzag_node *self, double x, double y)
+{
+  UNUSED(self);
+  UNUSED(x);
+  UNUSED(y);
+  zn_warn("VR Mode starting...");
```
Aki-7, line 59:
```suggestion
  zn_debug("Logging out...");
```
or 
```suggestion
  // TODO: Logout
```

zen/meson.build
```diff
@@ -42,15 +43,18 @@ _zen_srcs_pub = [
...
   m_dep,
   pixman_dep,
+  rsvg_dep,
   toml_dep,
   wayland_server_dep,
   wlroots_dep,
   wlr_glew_renderer_dep,
   xkbcommon_dep,
   zen_common_dep,
+  zigzag_dep,
```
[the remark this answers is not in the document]
liwii: Yes...

zen/screen/zigzag-layout.c
liwii: Fixed, the comment on the `clock_gettime` review gives the reason of the current design.

zen/screen/zigzag-layout.c
liwii: The "system clock" here does not show the time elapsed in the process but shows the real time clock, so the timer update should happen based on the `CLOCK_REALTIME` in that sense, the time going back and forth on NTP sync is completely reasonable. I keep using `CLOCK_REALTIME`, but changed the interval of the timer to a second so that the time change based on the NTP clock happens immediately. 

I also made a change to re-calculate the time the timer is fired in every handler call so that `CLOCK_REALTIME` change does not cause a cumulative problem.

meson.build
```diff
@@ -114,18 +114,28 @@ wlr_glew_renderer_dep = dependency('wlr-glew-renderer', version: wlr_glew_render
...
+rsvg_dep = dependency('librsvg-2.0')
 
 if get_option('cui-client')
   wayland_client_dep = dependency('wayland-client', version: wayland_req)
 endif
 
+icon_files = files(
+  'assets/icon/power-button.svg',
+)
+icon_install_dir = join_paths(datadir, 'icon', 'zen')
```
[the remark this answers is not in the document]
liwii: I completely agree with your point. Fixed.

liwii:
@Aki-7
Rewrote the code to follow the coding convention.

liwii:
I also moved `zn_zigzag_layout` from `zn_output` to `zn_screen` because in the current implementation `zn_default_cursor_grab` does not access `zn_output`.

However, the newly added datapath seems weird to me, so I want your opinion to fix.

zen/screen/output.c
```diff
@@ -148,26 +141,30 @@ zn_output_create(struct wlr_output *wlr_output)
...
+    goto err_damage;
   }
 
-  return self;
+  self->screen = zn_screen_create(
+      &screen_implementation, self, self->wlr_output, self->damage);
+  if (self->screen == NULL) {
+    zn_error("Failed to create z zn_screen");
+    goto err_damage;
+  }
```
Aki-7, line 165: `wl_signal_add`s called decades line above should after this because they also need to undo on error.

zen/screen.c
```diff
@@ -99,15 +102,25 @@ zn_screen_create(
...
+err_screen:
+  free(self);
 err:
   return NULL;
 }
 
 void
 zn_screen_destroy(struct zn_screen *self)
 {
+  zn_zigzag_layout_destroy(self->zn_zigzag_layout);
```
Aki-7, line 123: Please add this after 
```c
  wl_signal_emit(&self->events.destroy, NULL);
```

Emitting destroy signal should be first as signal handlers cannot assume the zn_screen is partially destroyed.
`zn_screen_layout_remove` is an exception.

zigzag/src/node.c
```diff
@@ -0,0 +1,72 @@
...
+  wl_list_init(&self->children);
+  self->frame = zalloc(sizeof self->frame);
+  self->implementation->set_frame(
+      self, layout->output_width, layout->output_height);
+
+  self->texture = zigzag_node_render_texture(self, renderer);
+
+err:
+  return self;
+}
```
Aki-7, line 62:
```suggestion
  self->texture = zigzag_node_render_texture(self, renderer);
  if (self->texture == NULL) { ... 
    goto err_free;
  }
  
  return self;

err_free:
  free(self);

err:
  return NULL;
}
```

zigzag/include/zigzag.h
```diff
@@ -0,0 +1,76 @@
...
+  void (*on_click)(struct zigzag_node *self, double x, double y);
+  void (*set_frame)(
+      struct zigzag_node *self, int output_width, int output_height);
+  void (*render)(struct zigzag_node *self, cairo_t *cr);
+};
+
+struct zigzag_node {
+  struct zigzag_layout *layout;
+
+  struct wlr_box *frame;
```
Aki-7, line 40:
```suggestion
  struct wlr_box frame;
```

zigzag/src/node.c
```diff
@@ -0,0 +1,72 @@
...
+
+void
+zigzag_node_destroy(struct zigzag_node *self)
+{
+  wl_list_remove(&self->children);
+  wlr_texture_destroy(self->texture);
+  if (self->user_data) free(self->user_data);
```
Aki-7, line 69: You don't need to care about the `user_data`.

zigzag/src/node.c
```diff
@@ -0,0 +1,72 @@
...
+  self->texture = zigzag_node_render_texture(self, renderer);
+
+err:
+  return self;
+}
+
+void
+zigzag_node_destroy(struct zigzag_node *self)
+{
+  wl_list_remove(&self->children);
```
Aki-7, line 67:
```suggestion
  wl_list_remove(&self->children);
  wl_list_remove(&self->link);
```

zigzag/src/node.c
```diff
@@ -0,0 +1,72 @@
...
+  if (self == NULL) {
+    zn_error("Failed to allocate memory");
+    goto err;
+  }
+
+  self->layout = layout;
+  self->user_data = user_data;
+  self->implementation = implementation;
+
+  wl_list_init(&self->children);
```
Aki-7, line 53:
```suggestion
  wl_list_init(&self->children);
  wl_list_init(&self->link);
```

zigzag/include/zigzag.h
```diff
@@ -0,0 +1,76 @@
...
+  struct wlr_texture *texture;
+
+  void *user_data;
+
+  struct wl_list link;  // for zigzag_layout :: nodes
```
Aki-7, line 45:
```suggestion
  struct wl_list link;  // for zigzag_layout::nodes or zigzag_node::child_list
```

Aki-7, line 45:
```suggestion
  struct wl_list link;  // for zigzag_layout::nodes or zigzag_node::node_list
```

zigzag/include/zigzag.h
```diff
@@ -0,0 +1,76 @@
...
+  struct wl_list children;
```
Aki-7, line 46:
```suggestion
  struct wl_list node_list;
``` 
We preffer `*_list` for `wl_list`, and `*s` for `wl_array`.

Aki-7, line 46:
```suggestion
  struct wl_list node_list;
``` 
We preffer `*_list` for `wl_list`, and `*s` for `wl_array`.

zen/ui/nodes/power-button.c
```diff
@@ -0,0 +1,168 @@
...
+
+  long time_ms = current_realtime_clock_ms();
+  self->next_sec_ms = (time_ms - time_ms % MSEC_PER_SEC + MSEC_PER_SEC);
+
+  int ms_delay = (int)(self->next_sec_ms - time_ms + 10);
+  wl_event_source_timer_update(self->second_timer_source, ms_delay);
+
+  zigzag_node->texture =
+      zigzag_node_render_texture(zigzag_node, server->renderer);
+  zigzag_node->layout->implementation->on_damage(zigzag_node);
```
Aki-7, line 26: Here, memory is leaked. zigzag_node already has a memory-allocated texture.
What about creating `zigzag_node_update_texture` which is equivalent to these three lines of code, and memory safe.

Aki-7: zigzag_node_render_texture will be static function then.
In general, assigning value outside the object's methods is potentially dangerous.

zen/ui/nodes/power-button.c
```diff
@@ -0,0 +1,168 @@
...
+  cairo_fill_preserve(cr);
+  cairo_set_line_width(cr, 0.5);
+  cairo_set_source_rgb(cr, 0.07, 0.12, 0.30);
+  cairo_stroke(cr);
+
+  FT_Library library;
+  FT_Init_FreeType(&library);
+  FT_Face face;
+  FT_New_Face(
+      library, "/usr/share/fonts/truetype/ubuntu/Ubuntu-R.ttf", 0, &face);
```
Aki-7, line 54: Is specifying full path like this safe? Or this should be fixed in the future?

zen/ui/nodes/power-button.c
```diff
@@ -0,0 +1,168 @@
...
+zn_power_button_on_click(struct zigzag_node *self, double x, double y)
+{
+  UNUSED(self);
+  UNUSED(x);
+  UNUSED(y);
+  // TODO: Show "Logout"
+}
+
+static void
+zn_power_button_render(struct zigzag_node *self, cairo_t *cr)
```
Aki-7, line 40: Following resources in this function doesn't seem to be destroyed properly
- `cr_face`
- `file`

zen/ui/nodes/menu-bar.c
```diff
@@ -0,0 +1,98 @@
...
+{
+  UNUSED(self);
+  UNUSED(x);
+  UNUSED(y);
+}
+
+static void
+zn_menu_bar_render(struct zigzag_node *self, cairo_t *cr)
+{
+  UNUSED(self);
```
Aki-7, line 21:
```suggestion
```

zen/ui/nodes/menu-bar.c
```diff
@@ -0,0 +1,98 @@
...
+  free(self);
+
+err:
+  return NULL;
+}
+
+void
+zn_menu_bar_destroy(struct zn_menu_bar *self)
+{
+  zn_power_button_destroy(self->power_button);
```
Aki-7, line 95: `self->power_button` is not set in the constructor, which leads to SIGSEGV on termination.

Aki-7: The same applies to `menu_bar` for `zn_zigzag_layout`.

zen/ui/nodes/power-button.c
Aki-7: It's ok to leave this as it is for now, but I want to know if there is any best practice to find font file.

zen/ui/nodes/power-button.c
liwii: Oh, that's true... I'll try to pack the update of memory-allocated attributes in an object by a setter method as much as possible.

zen/ui/nodes/menu-bar.c
liwii: Oh I totally forgot to do that, maybe I got satisfied after I set up the wl_list. Thanks for finding it out!!

zen/ui/nodes/power-button.c
```diff
@@ -0,0 +1,168 @@
...
+      cr, output, self->frame->width, self->frame->height, padding);
+
+  GError *error = NULL;
+  GFile *file = g_file_new_for_path(POWER_BUTTON_ICON);
+  RsvgHandle *handle = rsvg_handle_new_from_gfile_sync(
+      file, RSVG_HANDLE_FLAGS_NONE, NULL, &error);
+  if (handle == NULL) {
+    zn_error("Failed to create the svg handler: %s", error->message);
+    goto err;
+  }
```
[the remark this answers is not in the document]
liwii: I finally succeeded to load the svg only once in a load time, but it was more difficult than I thought.

zen/ui/nodes/power-button.c
liwii: I made a change to let `zigzag_layout` the font face as `system_font` attribute. I added some error handling with it.

liwii:
Updated the code to
- Use the continuous effective coordinate
- Hide the implementation of `zn_output` from `zn_zigzag_layout`

watasuke102 (approved):
LGTM!

zigzag/include/zigzag.h
```diff
@@ -0,0 +1,91 @@
...
+    cairo_t *cr, double width, double height, double radius);
+
+void zigzag_cairo_draw_centered_text(
+    cairo_t *cr, char *text, double width, double height);
+
+void zigzag_cairo_draw_left_aligned_text(
+    cairo_t *cr, char *text, double width, double height, double padding);
+
+void zigzag_cairo_draw_right_aligned_text(
+    cairo_t *cr, char *text, double width, double height, double padding);
```
watasuke102, line 85: Why do they require width and height? I thinl they should require x and y.

watasuke102 (commented):
Sorry to comment after approving, but I'm trying to use zigzag, and I have a question.

zigzag/include/zigzag.h
liwii: In the current design, `render()` gets a `cairo_t *`, not the surface itself. The `cairo_t*` is like a pen to draw lines on the surface, but it does not know the size of the surface itself.

Therefore, if you want to draw the text on the "center" of the surface, the caller needs to give the information about the width & height of the surface to the function.

If you want to draw the text at an arbitrary point of the surface, you can directly call `cairo_show_text` as what those utility functions do internally.

zen/ui/nodes/power-button.c
```diff
@@ -0,0 +1,169 @@
...
+      cr, self->frame.width, self->frame.height, self->frame.height / 2);
+  cairo_fill_preserve(cr);
+  cairo_set_line_width(cr, 0.5);
+  cairo_set_source_rgb(cr, 0.07, 0.12, 0.30);
+  cairo_stroke(cr);
+
+  time_t rawtime;
+  struct tm *timeinfo;
+
+  char output[8];
```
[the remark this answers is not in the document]
liwii: I just tried to pick the right size from power of 2 in the prototyping phase and left it. There is no much reason for 8 chars. Thanks for pointing it out.

zigzag/src/node.c
```diff
@@ -0,0 +1,102 @@
...
+  cairo_surface_destroy(surface);
+
+  return NULL;
+}
+
+static struct wlr_texture *
+zigzag_node_render_texture(
+    struct zigzag_node *self, struct wlr_renderer *renderer)
+{
+  cairo_surface_t *surface = zigzag_node_render_cairo_surface(self,
```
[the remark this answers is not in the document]
liwii: Thanks, added a line to destroy the surface before returning the texture,

zigzag/src/node.c
```diff
@@ -0,0 +1,102 @@
...
+
+  self->layout = layout;
+  self->user_data = user_data;
+  self->implementation = implementation;
+
+  wl_list_init(&self->node_list);
+  self->implementation->set_frame(
+      self, layout->screen_width, layout->screen_height);
+
+  self->texture = zigzag_node_render_texture(self, renderer);
```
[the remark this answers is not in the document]
liwii: Yeah true, I added the null-check and "// nonnull" comment.

zen/ui/nodes/power-button.c
```diff
@@ -0,0 +1,169 @@
...
+
+err:
+  return NULL;
+}
+
+void
+zn_power_button_destroy(struct zn_power_button *self)
+{
+  wl_event_source_remove(self->second_timer_source);
+  zigzag_node_destroy(self->zigzag_node);
```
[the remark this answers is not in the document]
liwii: Yes it is needed... Fixed.

zen/screen/renderer.c
```diff
@@ -155,6 +158,30 @@ render_view(struct zn_output *output, struct zn_view *view,
...
+    struct wl_list *node_list, pixman_region32_t *screen_damage)
+{
+  struct zigzag_node *node;
+  wl_list_for_each (node, node_list, link) {
+    float matrix[9];
+    struct wlr_box transformed_box;
+    zn_output_box_effective_to_transformed_coords(
+        output, &node->frame, &transformed_box);
+    wlr_matrix_project_box(matrix, &transformed_box, WL_OUTPUT_TRANSFORM_NORMAL,
+        0, output->wlr_output->transform_matrix);
```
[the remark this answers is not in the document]
liwii: I inserted a few lines based on the implementation  of `render_cursor` or `render_view`. Tell me if there is anything wrong.

zen/ui/nodes/power-button.c
```diff
@@ -0,0 +1,169 @@
...
+  bool result = zigzag_cairo_stamp_svg_on_surface(
+      cr, POWER_BUTTON_ICON, 0., 0., icon_width, icon_height);
+  if (!result) {
+    return false;
+  }
+  return true;
+}
+
+static bool
+zn_power_button_render(struct zigzag_node *self, cairo_t *cr)
```
Aki-7, line 53: In this file, `self` is `zn_power_button`. Sorry for missing in previous review, but please fix it in several methods like so.
```suggestion
zn_power_button_render(struct zigzag_node *node, cairo_t *cr)
```

zen/ui/nodes/power-button.c
```diff
@@ -0,0 +1,169 @@
...
+  timeinfo = localtime(&rawtime);
+
+  sprintf(output, "%02d:%02d", timeinfo->tm_hour, timeinfo->tm_min);
+
+  double padding = 6.;
+  cairo_set_font_size(cr, 11);
+  zigzag_cairo_draw_text(cr, output, padding, self->frame.height / 2,
+      ZIGZAG_ANCHOR_LEFT, ZIGZAG_ANCHOR_CENTER);
+
+  struct zn_power_button *power_button = self->user_data;
```
Aki-7, line 78:
```suggestion
  struct zn_power_button *self = node->user_data;
```

zen/ui/nodes/power-button.c
```diff
@@ -0,0 +1,169 @@
...
+  cairo_set_source_rgba(cr, 1.0, 1.0, 1.0, 0.7);
+  zigzag_cairo_draw_rounded_rectangle(
+      cr, self->frame.width, self->frame.height, self->frame.height / 2);
+  cairo_fill_preserve(cr);
+  cairo_set_line_width(cr, 0.5);
+  cairo_set_source_rgb(cr, 0.07, 0.12, 0.30);
+  cairo_stroke(cr);
+
+  time_t rawtime;
+  struct tm *timeinfo;
```
Aki-7, line 64: Are `rawtime` and `timeinfo` frequently used without `_`?
```suggestion
  time_t raw_time;
  struct tm *time_info;
```

zen/ui/nodes/power-button.c
```diff
@@ -0,0 +1,169 @@
...
+{
+  UNUSED(self);
+  cairo_set_source_rgba(cr, 1.0, 1.0, 1.0, 0.7);
+  cairo_paint(cr);
+  bool result = zigzag_cairo_stamp_svg_on_surface(
+      cr, POWER_BUTTON_ICON, 0., 0., icon_width, icon_height);
+  if (!result) {
+    return false;
+  }
+  return true;
```
Aki-7, line 49:
```suggestion
  return result;
```

zen/ui/nodes/power-button.c
liwii: Thanks, I resolved all `zigzag_node *self` arguments in `zen/ui/nodes` to `node`.

Aki-7 (approved):
```
         ,.  '                 ,.-·^*ª'` ·,                 ,  . .,  °        ,·'´¨;.  '                   
       /   ';\              .·´ ,·'´:¯'`·,  '\‘        ;'´    ,   ., _';\'      ;   ';:\           .·´¨';\   
     ,'   ,'::'\           ,´  ,'\:::::::::\,.·\'       \:´¨¯:;'   `;::'\:'\    ;     ';:'\      .'´     ;:'\  
    ,'    ;:::';'         /   /:::\;·'´¯'`·;\:::\°       \::::;   ,'::_'\;'    ;   ,  '·:;  .·´,.´';  ,'::;'  
    ';   ,':::;'         ;   ;:::;'          '\;:·´           ,'  ,'::;'  ‘     ;   ;'`.    ¨,.·´::;'  ;:::;   
    ;  ,':::;' '        ';   ;::/      ,·´¯';  °             ;  ;:::;  °     ;  ';::; \*´\:::::;  ,':::;‘   
   ,'  ,'::;'           ';   '·;'   ,.·´,    ;'\               ;  ;::;'  ‘     ';  ,'::;   \::\;:·';  ;:::; '   
   ;  ';_:,.-·´';\‘    \'·.    `'´,.·:´';   ;::\'             ;  ;::;'‚       ;  ';::;     '*´  ;',·':::;‘     
   ',   _,.-·'´:\:\‘    '\::\¯::::::::';   ;::'; ‘           ',.'\::;'‚       \´¨\::;          \¨\::::;      
    \¨:::::::::::\';      `·:\:::;:·´';.·´\::;'               \::\:;'‚        '\::\;            \:\;·'       
     '\;::_;:-·'´‘            ¯      \::::\;'‚                \;:'      ‘      '´¨               ¨'         
       '¨                              '\:·´'                    °                                         
```

PR #285, opened 2022-12-30
set initial board pos

watasuke102:
## Context

We can change the number of initial board, but they are overlapped.

## Summary
Place board at the different pos on seat_capsule by `zns_seat_capsule_rearrange()`.

## How to check behavior

Edit `config.toml`, and start zen and with remote display system

### DIFF
--- a/zns/seat-capsule.c
+++ b/zns/seat-capsule.c
@@ -2,12 +2,35 @@
 
 #include <cglm/quat.h>
 #include <cglm/vec3.h>
+#include <math.h>
 #include <zen-common.h>
 
 #include "board.h"
 #include "bounded.h"
 #include "zen/appearance/board.h"
 #include "zen/virtual-object.h"
+
+#define INITIAL_BOARD_GAP (M_PI / 10.f)
+
+void
+zns_seat_capsule_rearrange(struct zns_seat_capsule *self)
+{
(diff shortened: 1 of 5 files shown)


zns/seat-capsule.c
```diff
@@ -107,7 +130,6 @@ zns_seat_capsule_add_board(
 {
   wl_list_insert(&self->board_list, &board->seat_capsule_link);
 
-  // TODO: calculate better initial position
```
Aki-7, line 110: I think it's not finished yet.

zns/seat-capsule.c
```diff
@@ -2,13 +2,36 @@
...
+
+void
+zns_seat_capsule_rearrange(struct zns_seat_capsule *self)
+{
+  int board_len = wl_list_length(&self->board_list);
+  int i = 0;
+  float prev_azim;
+  struct zns_board *board;
+  wl_list_for_each (board, &self->board_list, seat_capsule_link) {
+    float azim;
```
Aki-7, line 23:
```suggestion
    float azimuthal;
```

We do not prefer abbreviations unless they are widely accepted.
The same applies to `prev_azim`.

Although "board_len" is probably acceptable, I would use "board_length" because I don't want to think about whether I can use that abbreviation.

Aki-7 (approved):

PR #295, opened 2023-01-06
Show VR modal

watasuke102:
## Context

We want to show information like `You're in VR`

## Summary

Add `zn_vr_modal` using zigzag, and when the display system is switched, the modal is shown

## How to check behavior

Start zen, and remote display system. You can see the modal.


zen/screen/renderer.c
```diff
@@ -164,6 +164,10 @@ render_zigzag_nodes(struct zn_output *output, struct wlr_renderer *renderer,
 {
   struct zigzag_node *node;
   wl_list_for_each (node, node_list, link) {
+    if (!node->visible) {
```
liwii, line 167: Could you update the `visibility` in the constructor of `zn_power_menu_item_clock` and `zn_power_menu_item_logout` to make these component compatible with the new way of rendering?

zen/ui/nodes/vr-modal.c
```diff
@@ -0,0 +1,234 @@
...
+
+  cairo_set_source_rgba(cr, 1.0, 1.0, 1.0, 1.0);
+  zigzag_cairo_draw_text(cr, connection_status, center_x + 80, center_y,
+      ZIGZAG_ANCHOR_LEFT, ZIGZAG_ANCHOR_CENTER);
+
+  cairo_restore(cr);
+}
+
+static void
+zn_vr_modal_render_headset_dialog(struct zn_vr_modal *self, cairo_t *cr,
```
liwii, line 79: How about moving the headset dialog to a different struct like `zn_vr_modal_headset_status`?

include/zen/ui/zigzag-layout.h
```diff
@@ -6,11 +6,16 @@
 #include "zen/screen.h"
 #include "zen/ui/nodes/menu-bar.h"
 
+struct zn_vr_modal;
```
liwii, line 9: Could you include `vr-moda.h` here?

liwii (changes requested):

include/zen/ui/zigzag-layout.h
watasuke102: I think that's needless; other header (like `include/zen/cursor.h`) doesn't include header.

liwii: Got it, then could you remove the include of `menu-bar` and `screen` to align the style??

zen/ui/nodes/vr-modal.c
liwii: Ahh I see,

I think we can use `set_frame` every time `zigzag_node_update_texture` is called to realize a node which dynamically changing the size (sorry I should have implemented it in that way...)

In my opinion, even though the dialog itself does not have a click region, it is more natural to implement it as another struct.

I'll upload a small patch to the update_texture, could you try rewriting it with that small patch? (If you want to spend more time on the other parts of the project, I can work on it after this PR is merged!!)

liwii: #298

watasuke102: nice!

watasuke102:
@liwii I created `vr-modal` directory under `zen/ui/nodes/`. Do you want me to apply this change like this to `power-menu`? Or do you have any reason not to create subdirectory? (If so, I revert this change)

liwii:
Yeah there was no reason for the non-subdirectorty structure. If you could help me add the structure to `power-menu` that would be great!!

zen/remote.c
```diff
@@ -20,16 +20,18 @@ zn_remote_handle_new_peer(struct wl_listener *listener, void *data)
...
 
-  wl_signal_emit(&self->events.peer_list_changed, NULL);
-
   {  // FIXME: Do this when user clicks "Connect" button
     struct znr_session *session =
         znr_remote_create_session(self->znr_remote, peer->znr_remote_peer);
     if (session == NULL) return;
 
+    // if you process this block after clicking "Connect" button,
+    // you should emit `peer_list_changed` after setting session to zn_peer
```
Aki-7, line 29: I think the state of a peer (e.g. the peer has session or not) should be sent via like zn_peer::events::state_changed.

Aki-7 (commented):
I left an issue https://github.com/zigen-project/zen/issues/301

zen/screen/renderer.c
```diff
@@ -215,19 +219,21 @@ zn_screen_renderer_render(struct zn_output *output,
...
 
     render_zigzag_nodes(output, renderer,
         &output->screen->zn_zigzag_layout->zigzag_layout->node_list,
         &screen_damage);
   }
 
-  render_cursor(output, server->scene->cursor, renderer, &screen_damage);
+  if (server->display_system == ZN_DISPLAY_SYSTEM_SCREEN) {
+    render_cursor(output, server->scene->cursor, renderer, &screen_damage);
+  }
```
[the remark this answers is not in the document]
liwii: Yes, the `zigzag_node`s shouldn't have any dependencies on `board`

Aki-7 (commented):
one more!

Aki-7 (approved):
```
    ___       ___       ___       ___   
   /\__\     /\  \     /\  \     /\__\  
  /:/  /    /::\  \    \:\  \   /::L_L_ 
 /:/__/    /:/\:\__\   /::\__\ /:/L:\__\
 \:\  \    \:\:\/__/  /:/\/__/ \/_/:/  /
  \:\__\    \::/  /   \/__/      /:/  / 
   \/__/     \/__/               \/__/
```

PR #297, opened 2023-01-06
fix FT_Face destruction

Aki-7:
## Context

FT_Face created from a FT_Library will be destroyed when the FT_Library is destroyed.
Destroying both results in double free.

ref: https://freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_library

## Summary

Fix FT_Face destruction process.

### DIFF
--- a/zigzag/src/layout.c
+++ b/zigzag/src/layout.c
@@ -29,12 +29,6 @@ zigzag_layout_create(const struct zigzag_layout_impl *implementation,
   if (FT_New_Face(library, font_file_path, 0, &ft_face) == 0) {
     self->system_font = cairo_ft_font_face_create_for_ft_face(ft_face, 0);
     cairo_status_t status = cairo_font_face_set_user_data(self->system_font,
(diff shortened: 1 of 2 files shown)


zigzag/src/layout.c
```diff
@@ -29,12 +29,6 @@ zigzag_layout_create(const struct zigzag_layout_impl *implementation,
   if (FT_New_Face(library, font_file_path, 0, &ft_face) == 0) {
     self->system_font = cairo_ft_font_face_create_for_ft_face(ft_face, 0);
     cairo_status_t status = cairo_font_face_set_user_data(self->system_font,
-        &self->ft_font_face_key, ft_face, (cairo_destroy_func_t)FT_Done_Face);
```
liwii, line 32: I think `ft_font_face_key` is not used anymore, could you help me remove it?

liwii (changes requested):
Thanks for catching it!! Just one comment...

zigzag/src/layout.c
Aki-7: Thanks for pointing that out!

liwii (approved):

PR #306, opened 2023-01-07
Fontconfig

watasuke102:
## Context

In Ubuntu, `Ubuntu-R.ttf` is at `/usr/share/fonts/truetype/ubuntu`.

But in Arch Linux, it's at `/usr/share/fonts/ubuntu`.

We should absorb this enviromental difference.

## Summary

Get font file path by fontconfig

## How to check behavior

Start zen. UI components like clock are rendered correctly.

### DIFF
--- a/common/font.c
+++ b/common/font.c
@@ -1,0 +1,60 @@
+#include <fontconfig/fontconfig.h>
+#include <string.h>
+
+#include "zen-common.h"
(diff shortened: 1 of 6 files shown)


Aki-7:
replaced with https://github.com/zigen-project/zen/pull/312

PR #309, opened 2023-01-08
introduce xwayland

watasuke102:
## Context

We want to launch X client.

## Summary

Create `wlr_xwayland`.

## How to check behavior

Start zen and some X client (ex: VSCode, Chrome).


MalekiRe:
What would be necessary to push this PR along?

Aki-7:
Sorry for the delay. Actually, it's not that difficult to show the main view of X applications through XWayland (thanks to wlroots), but considering other tasks like popups and system UIs (like the board switcher), we have to be careful about what to do first. We are now planning about that. Please bear with us!

PR #312, opened 2023-01-09
font

Aki-7:
## Context

Introduce system level font configuration.

## Summary

Hold font faces at global level.

## How to check behavior

No behavioral change is made

### DIFF
--- a/common/font.c
+++ b/common/font.c
@@ -1,0 +1,205 @@
+#include "zen-common/font.h"
+
+#include <cairo-ft.h>
+#include <fontconfig/fontconfig.h>
+#include <stddef.h>
+
+#include "zen-common/log.h"
+#include "zen-common/util.h"
+
+static struct zn_font *font_singleton = NULL;
+
+static const char *font_list[ZN_FONT_TYPE_COUNT] = {
+    [ZN_FONT_REGULAR] = "Ubuntu:style=Regular",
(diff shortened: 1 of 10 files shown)


common/font.c
```diff
@@ -0,0 +1,205 @@
...
+
+#include "zen-common/log.h"
+#include "zen-common/util.h"
+
+static struct zn_font *font_singleton = NULL;
+
+static const char *font_list[ZN_FONT_TYPE_COUNT] = {
+    [ZN_FONT_REGULAR] = "Ubuntu:style=Regular",
+    [ZN_FONT_BOLD] = "Ubuntu:weight=200",
+};
```
Aki-7, line 15: Font is specified by "<family-name>:<property>=<value>".

See https://www.freedesktop.org/software/fontconfig/fontconfig-user.html for properties and values details

liwii: Isn't it possible to write “Ubuntu:bold” ? I took a look at the doc a little bit, and it seems like “bold” is equivalent to “weight=200”. See “Font Name” and “Configuration File Format” -> “<constant>”.

Aki-7: It can be. I'll check it later and fix it if possible. Thanks!

watasuke102 (approved):
Nice work!

common/font.c
Aki-7: It works, fixed to `Ubuntu:bold`. Thanks!

PR #326, opened 2023-01-13
exit Immersive Display System by keybind

watasuke102:
## Context

We want to leave Immersive Display System by pressing Meta+V

## Summary

Add keybind about leaving

## How to check behavior

Start zen, connect to zen-mirror, and press Meta+V


Aki-7 (approved):
LGTM!

PR #332, opened 2023-01-14
replace zigen -> zwin

watasuke102:
## Context

To coincide with a organization name change, replace `zigen` -> `zwin`

## How to check behavior

Build, install, and launch.


Aki-7 (approved):
LGTM!

